PR #104, opened 2026-03-06
feat: Session Info panel with agent config, stats, and controls

saltyskip:
## Summary
- Adds a **Session Info** tab (ℹ icon) to the agent detail right-pane toolbar
- Shows agent config (permissions, working dir, prompt), session stats (status, messages, cost), and session controls (kill process, delete session)
- Fixes the "no way to terminate a session" problem

### Session section
- Live status indicator (green=idle, yellow=busy, gray=stopped) with auto-refresh every 5s
- Message count, cumulative cost, busy-since timestamp
- Truncated session ID

### Agent section
- Agent name, working directory
- Permissions displayed as accent-colored tags (or "None (default-deny)")
- Prompt and system append previews (truncated)

### Actions section
- **Kill Process** — force-kills the Claude process (disabled when no process is alive)
- **Delete Session** — with confirmation step, triggers cleanup in the parent

## Files changed
- `SessionInfoPanel.tsx` (new) — the panel component
- `AgentDetailView.tsx` — add "info" tab to ReferenceTab type and toolbar
- `styles.css` — add `.session-info-*` styles

## Test plan
- [ ] Open an agent, click the ℹ icon in the right toolbar
- [ ] Verify session status shows correctly (idle/busy/stopped)
- [ ] Verify agent permissions display as tags
- [ ] Send a message, watch status update to "Busy" then back to "Idle"
- [ ] Click "Kill Process" while agent is busy — verify process stops
- [ ] Click "Delete Session" → confirm → verify session is removed

### DIFF
--- a/cthulu-studio/src/components/AgentDetailView.tsx
+++ b/cthulu-studio/src/components/AgentDetailView.tsx
@@ -1,9 +1,10 @@
-import { useState, useRef, useCallback, useMemo } from "react";
+import { useState, useRef, useCallback, useMemo, useEffect } from "react";
 import AgentChatView, { useAgentChat } from "./AgentChatView";
 import FileViewer from "./FileViewer";
 import DebugPanel from "./DebugPanel";
 import ChangesPanel from "./ChangesPanel";
 import SessionInfoPanel from "./SessionInfoPanel";
+import { getSessionStatus } from "../api/client";
 import type { PendingPermission, FileChangeData } from "../hooks/useGlobalPermissions";
 import type { DebugEvent } from "./chat/useAgentChat";
 
@@ -40,6 +41,21 @@ export default function AgentDetailView({
   const [filesFlex, setFilesFlex] = useState(1);
   const [referenceTab, setReferenceTab] = useState<ReferenceTab>("files");
   const containerRef = useRef<HTMLDivElement>(null);
+  const [infoBadge, setInfoBadge] = useState(0);
+
+  // Poll session status for info badge (message count)
+  useEffect(() => {
+    let cancelled = false;
+    const poll = async () => {
+      try {
+        const s = await getSessionStatus(agentId, sessionId);
+        if (!cancelled) setInfoBadge(s.message_count);
+      } catch { /* ignore */ }
+    };
+    poll();
+    const interval = setInterval(poll, 5000);
+    return () => { cancelled = true; clearInterval(interval); };
+  }, [agentId, sessionId]);
 
   // Derive hookChangedFiles from fileChanges filtered by this agent/session
   const hookChangedFiles = useMemo(() => {
@@ -158,6 +174,9 @@ export default function AgentDetailView({
             <svg width="14" height="14" viewBox="0 0 16 16" fill="currentColor">
               <path d="M8 1a7 7 0 1 0 0 14A7 7 0 0 0 8 1zm0 1a6 6 0 1 1 0 12A6 6 0 0 1 8 2zm-.5 3h1v1h-1V5zm0 2h1v5h-1V7z"/>
             </svg>
+            {infoBadge > 0 && referenceTab !== "info" && (
+              <span className="ref-toolbar-badge">{infoBadge}</span>
+            )}
           </button>
         </div>
       </div>
--- a/cthulu-studio/src/components/AgentDetailView.tsx
+++ b/cthulu-studio/src/components/AgentDetailView.tsx
@@ -3,10 +3,11 @@ import AgentChatView, { useAgentChat } from "./AgentChatView";
 import FileViewer from "./FileViewer";
 import DebugPanel from "./DebugPanel";
 import ChangesPanel from "./ChangesPanel";
+import SessionInfoPanel from "./SessionInfoPanel";
 import type { PendingPermission, FileChangeData } from "../hooks/useGlobalPermissions";
 import type { DebugEvent } from "./chat/useAgentChat";
 
-export type ReferenceTab = "files" | "changes" | "debug";
+export type ReferenceTab = "files" | "changes" | "debug" | "info";
 
 interface AgentDetailViewProps {
   agentId: string;
@@ -31,7 +32,7 @@ export default function AgentDetailView({
   onPermissionResponse,
   hookDebugEvents,
   onClearHookDebug,
-  onDeleted: _onDeleted,
+  onDeleted,
   fileChanges,
 }: AgentDetailViewProps) {
   const chat = useAgentChat(agentId, sessionId);
@@ -106,12 +107,18 @@ export default function AgentDetailView({
               gitSnapshot={chat.gitSnapshot}
               hookChangedFiles={hookChangedFiles}
             />
-          ) : (
+          ) : referenceTab === "debug" ? (
             <DebugPanel
               chatEvents={chat.debugEvents}
               hookEvents={hookDebugEvents}
               onClearChat={chat.clearDebugEvents}
               onClearHook={onClearHookDebug}
+            />
+          ) : (
+            <SessionInfoPanel
+              agentId={agentId}
+              sessionId={sessionId}
+              onSessionDeleted={onDeleted}
             />
           )}
         </div>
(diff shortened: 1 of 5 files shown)


saltyskip:

_Reviewing PR #104 — this may take a few minutes._

cthulu-studio/src/components/SessionInfoPanel.tsx
```diff
@@ -0,0 +1,230 @@
...
+  const [confirmDelete, setConfirmDelete] = useState(false);
+
+  // Fetch agent + session status on mount and periodically
+  useEffect(() => {
+    let cancelled = false;
+    const fetchAll = async () => {
+      try {
+        const [a, s] = await Promise.all([
+          getAgent(agentId),
+          getSessionStatus(agentId, sessionId),
```
saltyskip, line 95: Nit: `getAgent(agentId)` is re-fetched every 5 seconds alongside session status, but agent config (name, permissions, prompt) rarely changes — it only updates when the user edits the agent. Consider fetching the agent once on mount (or when `agentId` changes) and only polling `getSessionStatus` in the interval. This halves the polling traffic.

```ts
// Fetch agent once
useEffect(() => {
  getAgent(agentId).then(setAgent).catch(() => {});
}, [agentId]);

// Poll only session status
useEffect(() => {
  let cancelled = false;
  const poll = async () => { ... getSessionStatus only ... };
  poll();
  const interval = setInterval(poll, 5000);
  return () => { cancelled = true; clearInterval(interval); };
}, [agentId, sessionId]);
```

saltyskip (commented):
## 🤖 Cthulu Review

Clean PR. The new SessionInfoPanel is well-structured with proper cleanup patterns (cancelled flag + clearInterval), a confirmation step on delete, and correct disabled states on action buttons.

**What looks good:**
- Correctly wires up the previously-unused `onDeleted` prop instead of aliasing it as `_onDeleted`
- Proper `useEffect` cleanup with cancellation flag prevents state updates on unmounted components
- Delete confirmation UX prevents accidental session deletion
- Kill button correctly disables when no process is alive
- CSS follows existing project conventions (CSS variables, `color-mix` for hover states)

**One nit (non-blocking):**
- The 5s polling interval re-fetches both agent config *and* session status. Agent config is static — consider splitting into a one-time fetch + status-only polling to reduce unnecessary API calls. See inline comment.

**Verdict:** Approve ✅

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #104 (previous HEAD: `03247ad`, new HEAD: `49dceef`)_

cthulu-studio/src/components/SessionInfoPanel.tsx
```diff
@@ -0,0 +1,669 @@
...
+    } catch { setError("Failed to save permissions"); }
+    setEditingPerms(false);
+  };
+
+  const copySessionId = () => {
+    navigator.clipboard.writeText(sessionId);
+    setCopied(true);
+    setTimeout(() => setCopied(false), 2000);
+  };
+
```
saltyskip, line 277: **Bug: `onBlur={cancelEdit}` races with `saveEdit` on Enter**

When the user presses Enter, `handleEditKeyDown` calls `saveEdit()`, but the input also loses focus, triggering `onBlur={cancelEdit}`. Since `cancelEdit` resets `editingField` to `null` synchronously, it can race with (or cancel) the async `saveEdit` — the `saveEdit` check `if (!agent || !editingField) return` will bail because `cancelEdit` already cleared `editingField`.

Fix: use `onBlur={saveEdit}` instead of `onBlur={cancelEdit}`, or guard the blur handler:
```tsx
onBlur={() => {
  // Don't cancel if save is already in-flight
  // Simplest fix: just save on blur too
  saveEdit();
}}
```
Or use a ref to track whether a save was initiated and skip the cancel.

cthulu-studio/src/components/AgentDetailView.tsx
```diff
@@ -31,14 +33,29 @@ export default function AgentDetailView({
...
+  useEffect(() => {
+    let cancelled = false;
+    const poll = async () => {
+      try {
+        const s = await getSessionStatus(agentId, sessionId);
+        if (!cancelled) setInfoBadge(s.message_count);
+      } catch { /* ignore */ }
+    };
+    poll();
+    const interval = setInterval(poll, 5000);
```
saltyskip, line 56: **Duplicate polling**: This `useEffect` polls `getSessionStatus` every 5s just for the badge count. `SessionInfoPanel` also polls the same endpoint every 5s. When the info tab is active, that's 2 identical requests every 5 seconds.

Consider lifting the status into shared state, or having `SessionInfoPanel` report the message count up via a callback prop so this effect isn't needed.

cthulu-studio/src/components/SessionInfoPanel.tsx
```diff
@@ -0,0 +1,669 @@
...
+                if (child.type === "file") {
+                  skills.push(child.path);
+                }
+              }
+            }
+            if (e.type === "directory" && e.children) {
+              findSkills(e.children, e.path);
+            }
+          }
+        };
```
saltyskip, line 133: Unused variable: `rootNames` is computed but never read. Should be removed.

cthulu-studio/src/components/SessionInfoPanel.tsx
```diff
@@ -0,0 +1,669 @@
...
+    const discover = async () => {
+      try {
+        const data = await listSessionFiles(agentId, sessionId);
+        if (cancelled) return;
+
+        // Find .skills/ directory
+        const skills: string[] = [];
+        const findSkills = (entries: FileTreeEntry[], parentPath: string) => {
+          for (const e of entries) {
+            if (e.type === "directory" && e.name === ".skills" && e.children) {
```
saltyskip, line 122: Unused parameter: `parentPath` is declared but never used in `findSkills`. Should be removed or used.

cthulu-studio/src/components/SessionInfoPanel.tsx
```diff
@@ -0,0 +1,669 @@
...
+  const cancelEdit = () => { setEditingField(null); setEditValue(""); };
+
+  const saveEdit = async () => {
+    if (!agent || !editingField) return;
+    try {
+      const updates = editingField === "name"
+        ? { name: editValue }
+        : { working_dir: editValue || null };
+      const updated = await updateAgent(agentId, updates);
+      setAgent(updated);
```
saltyskip, line 241: **Folder ID collision**: The Skills folder item gets `id: \`skills/${skillsFiles[0]}\`` — the same ID as its first child. This means clicking the folder row could match the first child for active-state logic, and the `key` dedup (`${item.id}-folder` vs `${item.id}-item`) is the only thing preventing a React key collision.

Consider giving folder items a distinct ID like `"skills"` or `"skills-folder"` to avoid ambiguity:
```tsx
items.push({ id: "skills" as NavId, label: "Skills", depth: 0, isFolder: true, icon: "⚡" });
```

saltyskip (commented):
## 🤖 Cthulu Review

Nice feature — the session info panel fills a real gap (no way to inspect/kill sessions before). The layout reusing `fv-*` classes from FileViewer is a good call. A few things need fixing before merge:

### Must fix

1. **`onBlur={cancelEdit}` race condition** — Pressing Enter triggers both `saveEdit` (via keydown) and `cancelEdit` (via blur). The blur handler clears `editingField` before the async save completes, causing `saveEdit` to bail early. The edit silently fails. Use `onBlur={saveEdit}` or guard the blur.

2. **Duplicate polling** — `AgentDetailView` polls `getSessionStatus` every 5s for the badge, and `SessionInfoPanel` *also* polls the same endpoint every 5s. When the info tab is open, that's 2 identical requests every 5 seconds. Lift the status up or use a callback.

### Should fix

3. **Unused variables** — `rootNames` (line 133) and `parentPath` parameter (line 122) in SessionInfoPanel are dead code. Will likely trigger lint warnings.

4. **Folder nav ID collision** — Skills/Config folder items reuse the first child's ID (`skills/${skillsFiles[0]}`), which can cause confusing active-state behavior. Give folders their own distinct ID.

### Minor

5. The info badge shows `message_count` — is that the right metric for a badge? It never clears (messages only increase), so the badge will always show once you send a single message. Consider showing it only when there's a *change* since last viewed, or removing the badge entirely.

---

_Automated review by Cthulu_

PR #105, opened 2026-03-09
feat: Animated Agents/Flows tabs in sidebar using animejs createLayout

MPRR-Pandu:
## Summary
- Replace the separate Agents and Flows collapsible accordion sections with a VS Code-style tabbed interface
- Tab switching animated using anime.js v4 `createLayout` API (records positions, animates children to new positions)
- Prompts and Nodes sections remain as collapsible accordions below the tabs

## Changes
- **Sidebar.tsx**: Added `activeTab` state, `createLayout` setup/cleanup via refs, replaced two `<Collapsible>` sections with tab bar + single content area
- **styles.css**: Added `.sidebar-tab-bar`, `.sidebar-tab`, `.sidebar-tab.active`, `.sidebar-tab-content` — matches existing `.flow-workspace-tab` VS Code style (11px, uppercase, accent underline, no border-radius)

## Details
- `createLayout(ref, { duration: 300, ease: "out(3)" })` on the tab content container
- On tab click: `layout.record()` → `setActiveTab()` → `useEffect` calls `layout.animate()`
- `layout.revert()` on unmount for cleanup
- `+` button in tab bar creates agent or flow depending on active tab


saltyskip:

_Reviewing PR #105 — this may take a few minutes._

cthulu-backend/agent_sdk/session.rs
```diff
@@ -0,0 +1,141 @@
...
+
+            let mut stream = Box::pin(response_stream);
+
+            while let Some(result) = stream.next().await {
+                match result {
+                    Ok(msg) => {
+                        if captured_session_id.is_none() {
+                            captured_session_id = message::extract_session_id(&msg);
+                        }
+
```
saltyskip, line 80: **Bug: `send_message` drives the entire stream to completion before returning the receiver.** The `while let Some(result) = stream.next().await` loop runs to completion (until `is_result()` or error), then sends a `done` event at line 111, then returns `Ok(rx)`. By the time the caller gets the `Receiver`, all events have already been sent to the broadcast channel — and since there were no subscribers yet, every event is dropped.

The caller in `chat_sdk_stream` (chat.rs) does `match session.send_message(&prompt).await { Ok(mut rx) => { ... loop over rx ... } }` — but `rx` will never receive anything because the stream already finished.

You need to either:
1. Spawn the stream consumption in a background task and return `rx` immediately, or
2. Return the raw SDK stream and let the caller drive it

As written, the SDK path will produce no output to the SSE client.

cthulu-backend/api/agents/chat.rs
```diff
@@ -62,6 +68,14 @@ fn process_key(agent_id: &str, session_id: &str) -> String {
     format!("agent::{agent_id}::session::{session_id}")
 }
 
+/// Check if the Agent SDK integration is enabled via AGENT_SDK_ENABLED env var.
+/// When enabled, new chat sessions use `AgentSession` (SDK) instead of `LiveClaudeProcess`.
+fn agent_sdk_enabled() -> bool {
```
saltyskip, line 73: **`agent_sdk_enabled()` reads env var on every call.** This is called in multiple hot paths (list_sessions, session_status, the main chat handler). Consider caching this in `AppState` at startup or using `std::sync::LazyLock`.

CLAUDE.md
```diff
@@ -12,7 +12,7 @@ This guide is the entry point for AI assistants working in this monorepo. It pro
 4. **This file (`CLAUDE.md`)** — Project-specific context, architecture, and workflows
 5. **Relevant `.claude/skills/*.skill.md`** for the task domain
 
-> **Generic rules in `.claude/rules/` apply to ALL tasks as baseline defaults. Project-specific rules in this file take precedence when they conflict**, since they encode domain-specific constraints (e.g., React Flow node merging, Axum param syntax) that generic rules cannot anticipate.
+> **Generic rules in `.claude/rules/` apply to ALL tasks and override project-specific guidance when they conflict.**
```
saltyskip, line 15: **This reverses the rule precedence.** Previously project-specific rules took precedence over generic rules (which makes sense — domain-specific constraints like React Flow node merging should override generic advice). This change makes generic rules override project-specific guidance, which seems like it would weaken the project-specific guardrails documented in this very file. Was this intentional?

saltyskip (changes requested):
## 🤖 Cthulu Review

**This PR bundles multiple unrelated changes that should be separate PRs:**

1. **Sidebar tabs UI** (Sidebar.tsx + styles.css) — the stated purpose of this PR
2. **Agent SDK integration** (new `agent_sdk/` module, chat.rs changes, AppState additions) — a major backend feature
3. **Nx version bump 20 → 22** (package.json, package-lock.json) — a major dependency upgrade
4. **CLAUDE.md rule precedence reversal** — changes AI assistant behavior
5. **Lessons.md cleanup** — documentation trimming

### Blocking Issues

**`send_message` is broken** — The `AgentSession::send_message()` method drives the SDK response stream to completion *before* returning the broadcast receiver. Since there are no subscribers during the stream processing, all events are silently dropped. The SSE client will receive nothing when using the SDK path. This needs to be restructured so the stream is consumed in a spawned task while the receiver is returned immediately.

**Potential deadlock in `list_sessions`** — Both `live_processes` and `sdk_sessions` mutexes are acquired simultaneously. If any other code path acquires them in the opposite order, this deadlocks. The same pattern appears in `session_status`. Either establish and document a consistent lock ordering, or restructure to avoid holding both.

### Non-blocking Issues

- `agent_sdk_enabled()` reads the env var on every request — cache it at startup
- `send_message` sends a `done` event, then `chat_sdk_stream` sends another `done` — double `done` events to the client
- The Sidebar animation code uses `useEffect` to call `layout.animate()` on `activeTab` change. Per CLAUDE.md rule #2, `useEffect` shouldn't be used for derived state/side effects from state changes. Consider calling `animate()` directly in the click handler after `setActiveTab`.
- The `CLAUDE.md` rule precedence change (generic rules now override project-specific) seems like it would undermine the project-specific guardrails. Verify this is intentional.

### Recommendation

Split this into separate PRs:
1. Sidebar tabs UI (small, reviewable)
2. Agent SDK integration (needs the `send_message` bug fixed first)
3. Nx upgrade (test separately)

---

_Automated review by Cthulu_

PR #106, opened 2026-03-09
feat: Claude CLI shortcuts skill doc

MPRR-Pandu:
## Summary
- Add `.claude/skills/claude-cli-shortcuts.skill.md` — a skill doc covering 10 Claude CLI use cases for the Cthulu project
- Available via `/claude-cli-shortcuts` in Claude Code sessions

## Use Cases Covered
1. **One-shot code review** — `cat file | claude -p` for quick reviews
2. **Resume sessions** — `claude -r` / `claude -c` for session continuity
3. **Model selection** — `--model sonnet/opus` for cost vs capability tradeoffs
4. **Budget-capped automation** — `--max-turns` and `--max-budget-usd` for CI safety
5. **Custom subagents** — `--agents` JSON for project-specific agents (flow reviewer, rust checker)
6. **Hooks for auto-formatting** — PostToolUse hooks for prettier/cargo check after edits
7. **Parallel git worktrees** — `claude -w` for isolated parallel sessions
8. **Checkpointing** — `Esc+Esc` rewind for safe experimentation
9. **Slash commands** — `.claude/commands/` for project shortcuts
10. **Permission modes** — `--permission-mode` for read-only vs auto-approve workflows

## Context
Based on analysis of current Claude CLI usage across 4 spawn contexts in the backend (pipeline executor, sandbox executor, interactive chat, utility spawns). Includes Cthulu-specific notes on gaps where unused CLI features could be leveraged (e.g. `--model` not used anywhere, `--max-turns` absent from pipeline runs).


saltyskip:

_Reviewing PR #106 — this may take a few minutes._

cthulu-backend/agent_sdk/session.rs
```diff
@@ -0,0 +1,141 @@
...
+                        break;
+                    }
+                }
+            }
+        }
+
+        let _ = tx.send(SseEvent {
+            event_type: "done".to_string(),
+            data: serde_json::json!({}),
+        });
```
saltyskip, line 108: **Bug: `send_message` blocks until the entire stream is consumed before returning the receiver.**

This method awaits the full response stream (lines 80-108), sends all events to `tx`, then returns `rx`. By the time the caller gets the receiver, all events are already buffered in the broadcast channel. This means:

1. **No real-time streaming** — the client doesn't receive SSE events as they arrive; they all get flushed at once after the response completes.
2. **Buffer overflow risk** — if the response has >1024 events, earlier ones are dropped (`RecvError::Lagged`).

The stream consumption should happen in a spawned task so `send_message` returns immediately with the receiver:

```rust
pub async fn send_message(&mut self, prompt: &str) -> Result<broadcast::Receiver<SseEvent>> {
    // ... setup ...
    let (tx, rx) = broadcast::channel::<SseEvent>(1024);
    
    // Spawn stream processing in background
    let response_stream = client.receive_response()?;
    tokio::spawn(async move {
        let mut stream = Box::pin(response_stream);
        while let Some(result) = stream.next().await {
            // ... process and send to tx ...
        }
    });
    
    Ok(rx)  // Return immediately
}
```

This is the most critical issue — it completely negates the SSE streaming architecture.

cthulu-backend/api/agents/chat.rs
```diff
@@ -438,12 +492,302 @@ pub(crate) struct StopRequest {
...
 }
 
+fn build_sdk_config(
+    agent: &crate::agents::Agent,
+    session_id: &str,
+    working_dir: &str,
+    is_new: bool,
+    system_prompt: Option<&str>,
+) -> SessionConfig {
+    let permission_mode = if agent.permissions.is_empty() {
```
saltyskip, line 502: **Security: defaulting to `bypassPermissions` when the permissions list is empty is backwards.**

Empty `agent.permissions` means the user hasn't configured any permissions for this agent. Defaulting to `bypassPermissions` (auto-approve everything) is the most dangerous possible default. It should be `default` (ask for everything) or at most `acceptEdits`.

```rust
let permission_mode = if agent.permissions.is_empty() {
    Some("default".to_string())
} else {
    Some("default".to_string())
};
```

Or remove the branch entirely since both paths would be the same.

cthulu-backend/api/agents/chat.rs
```diff
@@ -438,12 +492,302 @@ pub(crate) struct StopRequest {
...
+    agent: crate::agents::Agent,
+) -> impl Stream<Item = Result<Event, Infallible>> {
+    async_stream::stream! {
+        let key_for_stream = agent_key(&id);
+        let proc_key = process_key(&id, &target_session_id);
+        let sessions_ref = state.interact_sessions.clone();
+        let sessions_path = state.sessions_path.clone();
+        let sdk_sessions = state.sdk_sessions.clone();
+        let session_streams = state.session_streams.clone();
+        let chat_event_buffers = state.chat_event_buffers.clone();
```
saltyskip, line 537: **TOCTOU race condition in session creation.**

The `needs_create` check drops the lock, then later re-acquires it to insert. Two concurrent chat requests for the same agent+session could both see `needs_create = true` and create duplicate sessions, with the second overwriting the first.

Hold the lock across the check-and-insert, or use an entry API pattern.

CLAUDE.md
```diff
@@ -12,7 +12,7 @@ This guide is the entry point for AI assistants working in this monorepo. It pro
 4. **This file (`CLAUDE.md`)** — Project-specific context, architecture, and workflows
 5. **Relevant `.claude/skills/*.skill.md`** for the task domain
 
-> **Generic rules in `.claude/rules/` apply to ALL tasks as baseline defaults. Project-specific rules in this file take precedence when they conflict**, since they encode domain-specific constraints (e.g., React Flow node merging, Axum param syntax) that generic rules cannot anticipate.
+> **Generic rules in `.claude/rules/` apply to ALL tasks and override project-specific guidance when they conflict.**
```
saltyskip, line 15: **This reverses the intended precedence and is likely unintentional.**

The old text said project-specific rules in CLAUDE.md take precedence over generic `.claude/rules/` — which makes sense because CLAUDE.md encodes domain constraints (React Flow merging, Axum param syntax) that generic rules can't anticipate.

The new text says generic rules override project-specific guidance. This would mean generic coding style rules could override critical project-specific rules like the React Flow node merging constraint. Was this change intentional?

package.json
```diff
@@ -22,9 +22,9 @@
     "graph": "nx graph"
   },
   "devDependencies": {
-    "@nx/next": "^20.8.4",
-    "@nx/vite": "^20.8.4",
-    "nx": "^20.8.4",
+    "@nx/next": "^22.5.4",
+    "@nx/vite": "^22.5.4",
+    "nx": "^22.5.4",
```
saltyskip, line 27: **Nx major version bump (20 → 22) is unrelated to this PR.**

This is a major semver jump that could introduce breaking changes to the build system. It should be in its own PR with testing, not bundled into a "CLI shortcuts skill doc" PR.

cthulu-backend/api/agents/chat.rs
```diff
@@ -62,6 +68,14 @@ fn process_key(agent_id: &str, session_id: &str) -> String {
...
 }
 
+/// Check if the Agent SDK integration is enabled via AGENT_SDK_ENABLED env var.
+/// When enabled, new chat sessions use `AgentSession` (SDK) instead of `LiveClaudeProcess`.
+fn agent_sdk_enabled() -> bool {
+    std::env::var("AGENT_SDK_ENABLED")
+        .map(|v| matches!(v.to_lowercase().as_str(), "1" | "true" | "yes"))
+        .unwrap_or(false)
+}
+
```
saltyskip, line 78: Minor: `agent_sdk_enabled()` reads the env var on every request. If this is intentional to allow toggling without restart, consider documenting that. Otherwise, read it once in `Config` and pass through `AppState`.

cthulu-backend/agent_sdk/session.rs
```diff
@@ -0,0 +1,141 @@
...
+            bail!("session is busy processing a previous message");
+        }
+
+        self.busy = true;
+
+        let client = self.client.as_mut().unwrap();
+
+        client
+            .query(prompt)
+            .await
```
saltyskip, line 61: **`busy` flag is not reset if the future is dropped (e.g., client disconnect).**

`self.busy` is set to `true` at line 61 but only reset at line 119. If the caller cancels the future between those points, `busy` stays `true` forever, permanently locking the session. Consider using a guard/RAII pattern to reset `busy` on early exit.

saltyskip (changes requested):
## 🤖 Cthulu Review

### Scope Mismatch
This PR is titled "Claude CLI shortcuts skill doc" but contains significantly more than that:
- **New `agent_sdk` Rust module** with 3 files (~420 LOC) integrating `claude-agent-sdk-rust`
- **New Cargo dependency** (`claude-agent-sdk-rust`)
- **~300 LOC of changes to `chat.rs`** for SDK-based chat streaming
- **AppState changes** (new `sdk_sessions` field)
- **Nx major version bump** (20 → 22)
- **CLAUDE.md precedence reversal**
- **Lessons.md cleanup**

The skill doc itself looks good. The backend SDK integration has blocking issues.

### Blocking Issues

1. **`send_message` doesn't stream** — It awaits the entire response before returning the broadcast receiver. All events are buffered then flushed at once, defeating SSE streaming. This needs to be restructured so stream consumption happens in a spawned task.

2. **`bypassPermissions` as default** — When `agent.permissions` is empty, the code defaults to `bypassPermissions`. This should be `default` (prompt for everything). Empty permissions ≠ bypass all permissions.

3. **CLAUDE.md precedence reversal** — Changes project-specific rules from "override generic rules" to "are overridden by generic rules". This could cause AI assistants to ignore critical project constraints like React Flow node merging rules.

### Non-blocking Issues

4. **TOCTOU race** in SDK session creation — lock is dropped between check and insert
5. **`busy` flag leak** — not reset if the future is cancelled
6. **Env var read on every request** — `agent_sdk_enabled()` could be a startup config
7. **Nx 20→22 major bump** should be its own PR

### Recommendation
Split this into at least two PRs:
1. The skill doc + lessons cleanup (ready to merge)
2. The Agent SDK integration (needs the streaming bug and security default fixed first)

---
_Automated review by Cthulu_

PR #107, opened 2026-03-09
feat: Voice chat for agents + remove Claude Agent SDK

MPRR-Pandu:
## Summary
- Add voice input (STT) and voice output (TTS) to agent chat using the browser's Web Speech API
- Remove `claude-agent-sdk-rust` dependency — all agent sessions now use Claude CLI (`LiveClaudeProcess`) exclusively

## Voice Chat
- **New file**: `useVoiceChat.ts` hook — encapsulates `SpeechRecognition` (mic → text) and `SpeechSynthesis` (text → speech)
- Mic toggle button in the agent chat composer (between input and send button)
- When voice mode is active: mic listens → transcribes → auto-sends → agent responds → TTS reads response → mic re-listens
- Code blocks stripped from TTS output (reading code aloud isn't useful)
- Visual indicators: pulsing red when listening, green when speaking, accent when voice mode on
- Gracefully hidden when `SpeechRecognition` is unsupported (e.g. Firefox)

## Agent SDK Removal (-1000 lines)
- Delete `cthulu-backend/agent_sdk/` (mod.rs, session.rs, config.rs, message.rs)
- Remove `claude-agent-sdk-rust` from Cargo.toml
- Remove `sdk_sessions` from AppState, main.rs init, and shutdown
- Remove all SDK branches from chat.rs (list_sessions, delete_session, stop_chat, session_status, kill_session, stale-busy detection, chat handler)
- Remove `AGENT_SDK_ENABLED` / `ANTHROPIC_AUTH_TOKEN` from .env.example
- CLI-based `LiveClaudeProcess` path is fully preserved and unchanged

## Verification
- `cargo check` passes (no errors, only pre-existing warnings)
- `npx nx build cthulu-studio` passes


saltyskip:

_Reviewing PR #107 — this may take a few minutes._

cthulu-studio/src/components/chat/useVoiceChat.ts
```diff
@@ -0,0 +1,157 @@
...
+    utterance.onend = () => {
+      setIsSpeaking(false);
+      // Auto-restart listening after TTS finishes
+      if (isVoiceMode) startListening();
+    };
+    utterance.onerror = () => setIsSpeaking(false);
+
+    window.speechSynthesis.cancel();
+    window.speechSynthesis.speak(utterance);
+  }, [isVoiceMode, startListening]);
```
saltyskip, line 110: **Bug: stale `isVoiceMode` closure** — `speak` captures `isVoiceMode` via its dependency array, but `utterance.onend` fires asynchronously (potentially seconds later). If the user toggles voice mode off while TTS is playing, `isVoiceMode` in this closure is still `true`, so `startListening()` will be called after the user turned voice off.

Fix: use a ref for the check:
```ts
const isVoiceModeRef = useRef(isVoiceMode);
isVoiceModeRef.current = isVoiceMode;

// then in utterance.onend:
utterance.onend = () => {
  setIsSpeaking(false);
  if (isVoiceModeRef.current) startListening();
};
```

cthulu-studio/src/components/chat/useVoiceChat.ts
```diff
@@ -0,0 +1,157 @@
...
+  }, []);
+
+  const startListening = useCallback(() => {
+    if (!SpeechRecognitionImpl || !isVoiceMode) return;
+
+    const rec = new SpeechRecognitionImpl();
+    rec.continuous = false;
+    rec.interimResults = true;
+    rec.lang = "en-US";
+
```
saltyskip, line 65: Same stale closure issue here — `startListening` captures `isVoiceMode` at creation time. If voice mode is toggled off between when `speak`'s `onend` fires and calls `startListening`, the guard `if (!SpeechRecognitionImpl || !isVoiceMode)` may see a stale `true` value.

Using a ref (as suggested in the `speak` comment) for `isVoiceMode` would fix both sites.

cthulu-studio/src/components/chat/useVoiceChat.ts
```diff
@@ -0,0 +1,157 @@
...
+    rec.onresult = (e: SpeechRecognitionEventLike) => {
+      let interim = "";
+      let final = "";
+      for (let i = e.resultIndex; i < e.results.length; i++) {
+        const t = e.results[i][0].transcript;
+        if (e.results[i].isFinal) {
+          final += t;
+        } else {
+          interim += t;
+        }
```
saltyskip, line 75: Nit: `final` is a reserved word in strict mode. This works in practice since it's a `let` binding, but it can confuse linters and readers. Consider renaming to `finalText` or `finalTranscript`.

cthulu-studio/src/components/chat/useVoiceChat.ts
```diff
@@ -0,0 +1,157 @@
...
+  const speak = useCallback((text: string) => {
+    if (!window.speechSynthesis) return;
+    // Strip code blocks — reading code aloud isn't useful
+    const cleaned = text.replace(/```[\s\S]*?```/g, "(code block)").replace(/`[^`]+`/g, "");
+    const utterance = new SpeechSynthesisUtterance(cleaned);
+    utterance.rate = 1.1;
+
+    utterance.onstart = () => setIsSpeaking(true);
+    utterance.onend = () => {
+      setIsSpeaking(false);
```
saltyskip, line 102: Stripping inline code (`` `backtick` ``) replaces it with empty string, which can produce awkward speech like "use  instead of forEach". Consider replacing with the code content itself (just strip the backticks) rather than removing it entirely:
```ts
.replace(/`([^`]+)`/g, "$1")
```

cthulu-studio/src/components/chat/AgentChatThread.tsx
```diff
@@ -115,6 +116,36 @@ export default function AgentChatThread({
...
+          text = content;
+        } else if (Array.isArray(content)) {
+          text = (content as Array<Record<string, unknown>>)
+            .filter((p) => p.type === "text" && p.text)
+            .map((p) => p.text as string)
+            .join(" ");
+        }
+        if (text) voice.speak(text);
+      }
+    }
```
saltyskip, line 145: The `useEffect` depends on `isDone`, `voice.isVoiceMode`, and `messages`, but `voice.speak` is also called inside and is not in the dependency array. More importantly, `voice` object identity likely changes every render (it's a plain object returned from the hook), so depending on `voice.isVoiceMode` is fine but calling `voice.speak` without it being a dep is inconsistent.

Also, this effect will re-run whenever `messages` changes (e.g. when a new token streams in while `isDone` is still false from a previous turn), which is harmless but wasteful. Consider adding `messages.length` instead of the full array to reduce re-runs, or memoizing the messages array in the parent.

cthulu-studio/src/components/chat/AgentChatThread.tsx
```diff
@@ -115,6 +116,36 @@ export default function AgentChatThread({
...
+      if (lastMsg && lastMsg.role === "assistant") {
+        const content = lastMsg.content;
+        let text = "";
+        if (typeof content === "string") {
+          text = content;
+        } else if (Array.isArray(content)) {
+          text = (content as Array<Record<string, unknown>>)
+            .filter((p) => p.type === "text" && p.text)
+            .map((p) => p.text as string)
+            .join(" ");
```
saltyskip, line 141: If the last message has empty `content` (e.g., tool-use-only response), `text` will be empty and no speech fires, which is correct. But if the agent response has multiple assistant messages at the end (e.g., after tool use), this only reads the very last one. Might miss the main text response if a tool_result is appended after it. Consider searching backwards for the last assistant message with non-empty text content.

cthulu-studio/src/components/chat/AgentChatThread.tsx
```diff
@@ -115,6 +116,36 @@ export default function AgentChatThread({
...
+        }
+        if (text) voice.speak(text);
+      }
+    }
+    prevIsDoneRef.current = isDone;
+  }, [isDone, voice.isVoiceMode, messages]);
+
```
saltyskip, line 148: Setting `value={voice.transcript || undefined}` on `ComposerPrimitive.Input` — does this component actually support a controlled `value` prop? If it's based on `@assistant-ui/react`, the Composer Input is typically uncontrolled. Passing `value` may have no effect, or worse, conflict with internal state management. Please verify this works as expected.

CLAUDE.md
```diff
@@ -12,7 +12,7 @@ This guide is the entry point for AI assistants working in this monorepo. It pro
 4. **This file (`CLAUDE.md`)** — Project-specific context, architecture, and workflows
 5. **Relevant `.claude/skills/*.skill.md`** for the task domain
 
-> **Generic rules in `.claude/rules/` apply to ALL tasks as baseline defaults. Project-specific rules in this file take precedence when they conflict**, since they encode domain-specific constraints (e.g., React Flow node merging, Axum param syntax) that generic rules cannot anticipate.
+> **Generic rules in `.claude/rules/` apply to ALL tasks and override project-specific guidance when they conflict.**
```
saltyskip, line 15: This inverts the documented priority: previously project-specific rules (this file) took precedence over generic rules, which makes sense since project rules encode domain-specific constraints (React Flow node merging, Axum param syntax, etc.). The new wording says generic rules override project-specific guidance. Is this intentional? This seems like it could cause issues — generic rules can't anticipate project-specific constraints.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Voice chat hook has stale closure bugs** that will cause unexpected behavior — mic restarts after the user toggles voice off, because `isVoiceMode` is captured by value in `useCallback` but checked asynchronously in `utterance.onend` and `startListening`. Fix by using a ref for `isVoiceMode` checks inside async callbacks.

### Must fix
- **Stale closure on `isVoiceMode`** in `speak` and `startListening` — use `isVoiceModeRef` pattern (same as `onTranscriptRef`)
- **Verify `ComposerPrimitive.Input` supports `value` prop** — if it's `@assistant-ui/react`, the Input is typically uncontrolled and may ignore `value`

### Should fix
- Inline code stripping removes content entirely (`` `foo` `` → "") — better to just strip backticks
- `CLAUDE.md` rule priority inversion looks unintentional and potentially harmful

### Looks good
- Backend `BoxSseStream` type alias is a clean way to handle the SDK removal
- Voice hook structure and cleanup logic (unmount, toggle off) are solid
- CSS animations and visual states are well done
- Lessons.md condensation is welcome

---
_Automated review by Cthulu_

PR #110, opened 2026-03-10
feat: native sub-agent orchestration via --agents flag

MPRR-Pandu:
## Summary

- Replace standalone Looney Tunes agents (separate sessions each) with **sub-agent definitions** embedded in the Studio Assistant
- Use Claude Code's native `--agents` flag for single-session delegation — one session orchestrates code-reviewer, bugs-bunny, daffy-duck, and tweety-bird as sub-agents
- Add new **code-reviewer** sub-agent with structured review methodology: severity tagging (🔴/🟡/🟣), verification requirements, and REVIEW.md awareness (ported from the `automate_arepo` hat system)

## Architecture

```
Studio Assistant (parent session)
  └── --agents JSON
      ├── code-reviewer  (structured review, Read/Grep/Glob/Bash)
      ├── bugs-bunny     (personality reviewer, Read/Grep/Glob)
      ├── daffy-duck     (bug fixer, Read/Edit/Grep/Glob)
      └── tweety-bird    (test runner, Read/Bash/Glob)
```

Claude Code handles delegation natively — the parent agent decides when to spin up a sub-agent based on the task.

## Files changed

| File | Change |
|------|--------|
| `agents/mod.rs` | `SubAgentDef` type, `SubAgents` map, `subagents` + `subagent_only` fields on Agent, `default_subagents()` builder |
| `agents/code_reviewer_prompt.md` | New structured review prompt with severity system |
| `api/agents/chat.rs` | Pass `--agents <json>` to Claude CLI when agent has subagents |
| `api/agents/handlers.rs` | `subagent_only` + `subagent_count` in list API, Create/Update accept new fields |
| `main.rs` | Simplified seeding — only Studio Assistant (carries sub-agents) |
| Studio `types/flow.ts` | `subagent_only`, `subagent_count` on AgentSummary |
| Studio `Sidebar.tsx` | Filter `subagent_only` agents from sidebar |
| Studio `AgentGridView.tsx` | Filter `subagent_only` agents from grid |

## Test plan

- [ ] `cargo check` passes
- [ ] `npx nx build cthulu-studio` passes
- [ ] Delete `~/.cthulu/agents/` and restart server — Studio Assistant re-seeds with 4 sub-agents
- [ ] Open Studio chat — only Studio Assistant visible, no standalone Looney Tunes agents
- [ ] Send a task like "review src/main.rs, fix any bugs, then run tests" — verify Claude delegates to sub-agents within one session


saltyskip:

_Reviewing PR #110 — this may take a few minutes._

cthulu-backend/agent_sdk/session.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+        self.busy = false;
+
+        Ok(rx)
+    }
+
+    pub fn is_connected(&self) -> bool {
+        self.client
+            .as_ref()
+            .map_or(false, |c| c.is_connected())
```
saltyskip, line 121: 🔴 **`send_message` blocks until the entire response is consumed — no real-time streaming**

This method drives the SDK stream to completion (the `while let Some(result) = stream.next().await` loop processes every message), then sends a `done` event, and *then* returns `rx`. By the time the caller gets the receiver, all events have already been buffered in the broadcast channel.

In `chat_sdk_stream`, the spawned task calls `session.send_message().await`, which blocks for the entire response duration (could be minutes for a complex task). Only after it returns does the task start forwarding events to the outer broadcast channel (`bc_tx`). The user sees nothing until the full response is complete — defeating the purpose of SSE streaming.

**Fix**: Split this into two phases: (1) send the query and return the receiver immediately, (2) let the caller drive the stream. Or restructure so the SDK stream is consumed in the spawned task directly rather than going through a double-broadcast-channel indirection.

cthulu-backend/api/agents/chat.rs
```diff
@@ -438,12 +492,302 @@ pub(crate) struct StopRequest {
...
+                    // Remove old disconnected session if present
+                    pool.remove(&proc_key);
+                    pool.insert(proc_key.clone(), session);
+
+                    yield Ok(Event::default().event("system").data(
+                        serde_json::to_string(&json!({"message": "Session started (SDK)"})).unwrap()
+                    ));
+                }
+                Err(e) => {
+                    tracing::error!(error = %e, "failed to create AgentSession");
```
saltyskip, line 571: 🔴 **`sdk_sessions` mutex held across long-running `.await` — blocks all other sessions**

The mutex is acquired at the top (`let mut pool = sdk_sessions.lock().await`), and `session.send_message(&prompt).await` runs while the lock is held. Since `send_message` blocks until the entire response is consumed (see session.rs comment), this mutex is held for potentially minutes.

During that time, **all** other operations that need `sdk_sessions` are blocked: status checks, session creation, disconnect, kill, and other chat requests. This effectively serializes all SDK chat to one-at-a-time.

**Fix**: Take the session out of the pool (`pool.remove(&proc_key)`), drop the lock, use the session, then re-insert it after completion. Or use a per-session lock rather than a global pool lock.

cthulu-backend/api/agents/chat.rs
```diff
@@ -438,12 +492,302 @@ pub(crate) struct StopRequest {
...
+    is_new: bool,
+    working_dir: String,
+    system_prompt: Option<String>,
+    agent: crate::agents::Agent,
+) -> impl Stream<Item = Result<Event, Infallible>> {
+    async_stream::stream! {
+        let key_for_stream = agent_key(&id);
+        let proc_key = process_key(&id, &target_session_id);
+        let sessions_ref = state.interact_sessions.clone();
+        let sessions_path = state.sessions_path.clone();
```
saltyskip, line 534: 🔴 **SDK path doesn't pass sub-agents — the main feature of this PR is broken when `AGENT_SDK_ENABLED=1`**

`build_sdk_config` has no `subagents` field, and `SessionConfig` doesn't include them. The `--agents` flag is only added in the legacy CLI path (line ~1250). So when the SDK path is active, the Studio Assistant's sub-agents are silently ignored — Claude won't know about code-reviewer, bugs-bunny, etc.

Either add sub-agent support to the SDK config, or document that the SDK path doesn't support sub-agents yet.

cthulu-backend/api/agents/chat.rs
```diff
@@ -438,12 +492,302 @@ pub(crate) struct StopRequest {
...
+            !pool.contains_key(&proc_key) || !pool.get(&proc_key).map_or(false, |s| s.is_connected())
+        };
+
+        if needs_create {
+            let config = build_sdk_config(
+                &agent,
+                &target_session_id,
+                &working_dir,
+                is_new,
+                system_prompt.as_deref(),
```
saltyskip, line 552: 🟡 **TOCTOU race in `needs_create` check**

The lock is acquired to check `needs_create`, then dropped. Between the check and the subsequent `AgentSession::create` + re-acquire, another concurrent request could create a session for the same key. The old session would then be silently replaced (line 575), potentially leaking a connected process.

**Fix**: Hold the lock through the check-and-insert, or use an entry API pattern.

cthulu-backend/main.rs
```diff
@@ -149,19 +145,15 @@ async fn run_server(start_disabled: bool) -> Result<(), Box<dyn Error>> {
...
-        }
+    // Seed the built-in Studio Assistant (with sub-agents) if it doesn't exist yet.
+    // Sub-agents (code-reviewer, bugs-bunny, daffy-duck, tweety-bird) are embedded
+    // in the Studio Assistant's subagents field and passed via --agents to Claude Code.
+    if agent_repo.get(STUDIO_ASSISTANT_ID).await.is_none() {
+        tracing::info!("seeding built-in Studio Assistant agent with sub-agents");
+        agent_repo
+            .save(default_studio_assistant())
+            .await
+            .with_context(|| "failed to seed Studio Assistant agent")?;
```
saltyskip: 🟡 **Old standalone agents not cleaned up on migration**

This only seeds the Studio Assistant if it doesn't exist, but doesn't remove the old standalone agents (bugs-bunny, daffy-duck, tweety-bird) that were previously seeded as top-level agents. After upgrading, those agents will still exist on disk and still show in the UI (since `subagent_only` defaults to `false`). Users will see duplicate agents — the old standalone ones plus the new sub-agents inside Studio Assistant.

Consider adding a migration step that removes or marks the old agents as `subagent_only`.

cthulu-backend/api/agents/chat.rs
```diff
@@ -62,6 +68,14 @@ fn process_key(agent_id: &str, session_id: &str) -> String {
     format!("agent::{agent_id}::session::{session_id}")
 }
 
+/// Check if the Agent SDK integration is enabled via AGENT_SDK_ENABLED env var.
+/// When enabled, new chat sessions use `AgentSession` (SDK) instead of `LiveClaudeProcess`.
+fn agent_sdk_enabled() -> bool {
```
saltyskip, line 73: 🟡 **`agent_sdk_enabled()` re-reads the env var on every request**

This calls `std::env::var` on every chat request. Consider using `std::sync::OnceLock` or `std::sync::LazyLock` to read it once.

cthulu-backend/api/agents/chat.rs
```diff
@@ -90,8 +105,11 @@ pub(crate) async fn list_sessions(
             .iter()
```
saltyskip, line 105: 🟡 **Potential deadlock: `live_processes` and `sdk_sessions` locked simultaneously**

In `list_sessions`, both `state.live_processes.lock().await` (line 97) and `state.sdk_sessions.lock().await` (line 105) are held at the same time. If another code path acquires these locks in the opposite order, you'll get a deadlock. Ensure consistent lock ordering across all call sites, or restructure so both aren't held simultaneously.

saltyskip (changes requested):
## 🤖 Cthulu Review

The sub-agent architecture refactor (standalone agents → embedded sub-agents via `--agents` flag) is a solid design choice — single-session orchestration is cleaner than managing 4 separate Claude processes.

However, there are **3 blocking issues** in the new Agent SDK integration that need to be fixed before merging:

1. **No real-time streaming in SDK path** — `AgentSession::send_message()` blocks until the entire response is consumed before returning the receiver. Users won't see any output until Claude finishes (could be minutes). The double-broadcast-channel architecture needs restructuring so events flow in real-time.

2. **Global mutex held during entire response** — `sdk_sessions` is locked while `send_message` runs, blocking all other SDK operations (other chats, status checks, disconnects) for the duration. This effectively serializes all SDK chat.

3. **Sub-agents not passed via SDK path** — The `--agents` flag is only added in the legacy CLI path. When `AGENT_SDK_ENABLED=1`, the Studio Assistant's sub-agents are silently dropped, making the main feature of this PR non-functional under the SDK.

The sub-agent definitions themselves (`agents/mod.rs`), the Studio filtering (`Sidebar.tsx`, `AgentGridView.tsx`), and the legacy CLI `--agents` pass-through all look correct. The code-reviewer prompt is well-structured. The API handler changes for create/update/list are clean.

Additionally: old standalone agents (bugs-bunny, daffy-duck, tweety-bird) won't be cleaned up on upgrade — they'll remain visible alongside the new embedded sub-agents.

---

_Automated review by Cthulu_

saltyskip (approved):

MPRR-Pandu:
## 🐰 Bugs Bunny's Code Review

*"Eh, what's up doc?"*

### 🔴 Blocking Issues

**1. SDK path silently drops `--agents` flag** (`chat.rs`)
The core feature of this PR (sub-agents) is completely absent when `AGENT_SDK_ENABLED=1`. The `--agents` JSON is only threaded through the legacy `LiveClaudeProcess` path; `SessionConfig` has no field for sub-agents.

**2. No migration for existing installs** (`main.rs`)
The seeding guard only creates a new Studio Assistant if one doesn't exist. Existing installations already have one on disk without `subagents`, so the feature ships dead-on-arrival. Needs a migration check: if agent exists but `subagents.is_empty()`, update it with `default_subagents()`.

**3. Nested mutex locks → deadlock hazard** (`chat.rs`)
`live_processes` lock is held while acquiring `sdk_sessions` lock in the stale-session check. If any other code path locks them in reverse order, deadlock. Violates the project's "single lock pattern" rule — acquire once, drain into locals, drop the lock, then process.

### 🟡 Normal Bugs

**4. Double "done" event** (`chat.rs` / `chat_sdk_stream`)
`AgentSession::send_message` already emits a `done` event, then the background task sends another one. The duplicate gets buffered and corrupts reconnection replay via `chat_event_buffers`.

**5. `send_message` returns `Receiver` after stream is fully consumed** (`session.rs`)
All events are driven to completion inside `send_message` before returning the `Receiver`. Works now due to broadcast channel buffering but is fragile under load. The stream should be driven in a `tokio::spawn` with the `Receiver` returned immediately.

### 🟣 Nits

**6. `maxTurns` serde rename** (`agents/mod.rs`)
The `#[serde(rename = "maxTurns")]` on `max_turns` may mismatch Claude Code's `--agents` schema if it expects snake_case. Needs verification.

**7. Nx 20 → 22 major bump** (`package.json`)
Two-major-version jump in a sub-agent PR violates the project's own CODE_QUALITY rule 7 ("Build tool major bumps: Stay conservative"). Should be a separate PR with build verification.

---
🤖 Review by Bugs Bunny via [Claude Code](https://claude.com/claude-code)

saltyskip:

_Re-reviewing PR #110 (previous HEAD: `7240c25`, new HEAD: `88709d9`)_

saltyskip (changes requested):
## 🤖 Cthulu Review

Good architectural direction — consolidating standalone agents into sub-agent definitions passed via `--agents` is cleaner than managing separate sessions. The legacy path implementation for `--agents` looks correct, and the migration logic in `main.rs` is solid. However, the new Agent SDK code path has several issues that need fixing before merge.

### 🔴 Blocking Issues

**1. `sdk_sessions` mutex held for entire conversation turn** (`agent_sdk/session.rs:119`, `api/agents/chat.rs` spawn block)

`send_message()` processes the **entire** Claude response stream synchronously before returning. In `chat_sdk_stream`, it's called while holding the `sdk_sessions` pool lock:

```rust
let mut pool = sdk_sessions.lock().await;      // lock acquired
if let Some(session) = pool.get_mut(&proc_key) {
    match session.send_message(&prompt).await { // awaits ENTIRE response
        Ok(mut rx) => {
            drop(pool);                         // lock released after response completes
```

This blocks ALL other SDK session operations (create, list, status, kill, stop) for the full duration of a Claude turn (30+ seconds to minutes). The legacy `LiveClaudeProcess` path avoids this by only holding the lock briefly to write to stdin.

**Fix**: Restructure `send_message` to return the broadcast receiver immediately and spawn the stream processing internally, so the pool lock can be dropped right after starting the message.

### 🟡 Should Fix

**2. SDK path ignores sub-agents** (`api/agents/chat.rs`, `build_sdk_config`)

`build_sdk_config` and `SessionConfig` have no field for sub-agent definitions. When `AGENT_SDK_ENABLED=1`, the agent's `subagents` map is never passed to the Claude SDK client. The legacy path correctly passes `--agents <json>`, but the SDK path silently drops them. This means the main feature of this PR (sub-agent orchestration) won't work when the SDK is enabled.

**Fix**: Add a `subagents`/`agents` field to `SessionConfig`, wire it through `into_sdk()`, and populate it in `build_sdk_config`.

**3. Race condition: `bc_tx.subscribe()` after `tokio::spawn`** (`api/agents/chat.rs`, `chat_sdk_stream`)

The background task is spawned first, then the outer stream subscribes to `bc_tx` afterward. If the spawned task sends events before the subscribe, those events are lost. In practice this is unlikely due to the SDK session lock acquisition, but it's not guaranteed under load.

**Fix**: Subscribe before spawning:
```rust
let mut rx = bc_tx.subscribe(); // subscribe FIRST
tokio::spawn(async move { ... }); // then spawn
```

**4. Double "done" event** (`agent_sdk/session.rs:111-114`, `api/agents/chat.rs` background task)

`send_message()` sends a `done` event via its internal broadcast channel (line 111-114). The background task loop breaks on `done`, then sends *another* `done` event. The client receives two `done` events, which could confuse the frontend.

**Fix**: Remove the `done` event from `send_message` (let the caller handle it), or don't send the second one in the background task.

**5. Holding both `live_processes` and `sdk_sessions` locks simultaneously** (`api/agents/chat.rs`, `list_sessions` and `session_status`)

Both locks are acquired without dropping the first. If any other code path acquires them in the opposite order, it deadlocks (hanging the tokio runtime, not panicking). This is fragile.

**Fix**: Document a consistent lock ordering, or restructure to avoid holding both — check `live_processes` first, drop it, then check `sdk_sessions`.

**6. Security hooks removed without equivalent enforcement** (`agents/mod.rs`)

The old standalone agents had `PreToolUse` hooks as defense-in-depth:
- Bugs Bunny: denied `Edit`/`Write` via hook
- Tweety Bird: blocked destructive bash commands (`rm -rf /`, `git push --force`)
- Daffy Duck: denied `Write` (no new files)

These are all gone. The sub-agent `tools` array restricts exposure, but there's no second layer of defense if Claude Code's `--agents` tool filtering has a bug. Consider whether this is an acceptable tradeoff or if equivalent guardrails should be added.

### Summary

The **legacy path** changes (sub-agent definitions, `--agents` flag, Studio filtering, migration) look good and are the core value of this PR. The **Agent SDK path** (behind `AGENT_SDK_ENABLED` flag, off by default) has significant issues — the mutex contention alone would make it unusable for concurrent sessions. If the SDK path is intended as a prototype/WIP, consider documenting it as such or gating it more clearly.

---

_Automated review by Cthulu_

MPRR-Pandu:
## 🐰 Bugs Bunny's Re-Review of PR #110

*"Eh, what's up doc? I'm back."*

### Previous Bugs — Scorecard

| # | Bug | Status |
|---|-----|--------|
| 1 | Double "done" event in SDK path | ❌ Still present |
| 2 | Nested mutex / deadlock hazard | ❌ Still present |
| 3 | SDK path drops `--agents` flag | ❌ Still present |
| 4 | No migration for existing installs | ✅ **Fixed** |
| 5 | `maxTurns` serde rename breaks deserialization | ❌ Still present |
| 6 | Nx 20→22 major bump | ❌ Still present (commit claims revert but `package.json` still shows `^22.5.4`) |
| 7 | `send_message` returns Receiver after stream consumed | ❌ Still present |

**1 of 7 previous bugs fixed.**

---

### 🔴 New Blocking Issues

**NEW A — Migration orphans old standalone Looney Tunes agents**
The migration adds sub-agents to the Studio Assistant but never deletes the old standalone `bugs-bunny`, `daffy-duck`, `tweety-bird` YAML files on disk. Those old agents don't have `subagent_only: true`, so they still appear in the UI — duplicating agents instead of replacing them.

**NEW B — `list_sessions` three-mutex chain**
`list_sessions` acquires `interact_sessions` read lock, then `live_processes`, then `sdk_sessions` — all held simultaneously. This widens the deadlock surface from BUG 2 into a second handler.

**NEW C — SDK `busy` flag never reset on `query()` failure**
If `client.query(prompt).await` returns `Err`, `send_message` returns early via `?` but never resets `self.busy = false`. The session is permanently stuck — all subsequent messages get "session is busy" forever.

### 🟡 New Nit

**NEW D — `[RECONNECT-DEBUG]` log spam**
~8 `tracing::info!` calls with `[RECONNECT-DEBUG]` prefix left in production code paths. Should be `tracing::debug!` or removed.

---

**Bottom line**: 1/7 previous bugs fixed, 4 new issues found. Not ready to merge, doc.

*That's all folks!*

---
🤖 Re-review by Bugs Bunny via [Claude Code](https://claude.com/claude-code)

PR #111, opened 2026-03-11
feat: Animated Agents/Flows tabs in sidebar

MPRR-Pandu:
## Summary
- Replace the separate Agents and Flows collapsible accordion sections with a VS Code-style tabbed interface
- Tab switching animated using animejs v4 `createLayout` API (records positions, animates children to new positions)
- Prompts and Nodes sections remain as collapsible accordions below the tabs

## Changes
- **Sidebar.tsx**: Added `activeTab` state, `createLayout` setup/cleanup via refs, replaced two `<Collapsible>` sections with tab bar + single content area
- **styles.css**: Added `.sidebar-tab-bar`, `.sidebar-tab`, `.sidebar-tab.active`, `.sidebar-tab-content` — matches existing VS Code style (11px, uppercase, accent underline, no border-radius)

## Details
- `createLayout(ref, { duration: 300, ease: "out(3)" })` on the tab content container
- On tab click: `layout.record()` → `setActiveTab()` → `useEffect` calls `layout.animate()`
- `layout.revert()` on unmount for cleanup
- `+` button in tab bar creates agent or flow depending on active tab

## Test plan
- [ ] Verify Agents tab shows agent list with expand/collapse and sessions
- [ ] Verify Flows tab shows flow list with enable/disable switches
- [ ] Verify tab switching animates smoothly
- [ ] Verify `+` button creates agent on Agents tab, opens gallery on Flows tab
- [ ] Verify Prompts and Nodes accordion sections still work below tabs


saltyskip:

_Reviewing PR #111 — this may take a few minutes._

cthulu-studio/src/components/Sidebar.tsx
```diff
@@ -1,9 +1,10 @@
-import { useState, useEffect, useCallback } from "react";
+import { useState, useEffect, useCallback, useRef } from "react";
 import { STUDIO_ASSISTANT_ID, type FlowSummary, type Flow, type NodeTypeSchema, type AgentSummary, type SavedPrompt, type ActiveView } from "../types/flow";
 import { listAgents, createAgent, deleteAgent, listPrompts, savePrompt, deletePrompt as deletePromptApi, listAgentSessions, newAgentSession } from "../api/client";
 import type { InteractSessionInfo } from "../api/client";
 import { Switch } from "@/components/ui/switch";
 import { Collapsible, CollapsibleTrigger, CollapsibleContent } from "@/components/ui/collapsible";
+import { createLayout } from "animejs";
```
saltyskip, line 8: **Missing dependency**: `animejs` is not listed in `cthulu-studio/package.json` (neither `dependencies` nor `devDependencies`). This import will fail on a clean install. The PR needs to add `animejs` as a dependency — likely `"animejs": "^4.x"` in `dependencies`.

cthulu-studio/src/components/Sidebar.tsx
```diff
@@ -64,6 +65,20 @@ export default function Sidebar({
...
+  const tabContentRef = useRef<HTMLDivElement>(null);
+  const layoutRef = useRef<ReturnType<typeof createLayout> | null>(null);
+
+  useEffect(() => {
+    if (tabContentRef.current) {
+      layoutRef.current = createLayout(tabContentRef.current, { duration: 300, ease: "out(3)" });
+    }
+    return () => { layoutRef.current?.revert(); };
+  }, []);
+
```
saltyskip, line 79: This `useEffect` fires on mount too (before any `record()` has been called), calling `animate()` with no prior snapshot. Depending on animejs internals this is either a no-op or could cause a visual glitch on first render.

Consider guarding it:
```ts
useEffect(() => {
  if (layoutRef.current) {
    layoutRef.current.animate();
  }
}, [activeTab]);
```
…or better, skip the mount run entirely with a ref flag:
```ts
const hasSwitched = useRef(false);
useEffect(() => {
  if (hasSwitched.current) layoutRef.current?.animate();
  hasSwitched.current = true;
}, [activeTab]);
```

saltyskip, line 79: Per project rules (CLAUDE.md rule #2): prefer event handlers for side effects over `useEffect`. The `record()` call is already in the click handler — you could potentially move `animate()` there too using `flushSync` to force the synchronous re-render:
```tsx
import { flushSync } from 'react-dom';

onClick={() => {
  layoutRef.current?.record();
  flushSync(() => setActiveTab("agents"));
  layoutRef.current?.animate();
}}
```
This eliminates the `useEffect` entirely and makes the animation flow easier to follow. Worth trying — if the FLIP timing works with `flushSync`, it's cleaner.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Blocking: missing `animejs` dependency.** The PR imports `createLayout` from `animejs` but doesn't add it to `cthulu-studio/package.json`. This will break on any clean install or CI build.

Additionally:
- The animation `useEffect` fires on mount before any `record()` has been called — should guard against the initial render to avoid potential visual glitches
- Per project convention (CLAUDE.md rule #2), consider moving the `animate()` call into the click handler using `flushSync` instead of a separate `useEffect`, which would make the animation flow more explicit and eliminate a reactive dependency

The UI/CSS changes and the overall tab concept look solid — just needs the dependency fix and the minor animation timing cleanup.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #111 (previous HEAD: `5987289`, new HEAD: `03d0727`)_

cthulu-studio/src/components/Sidebar.tsx
```diff
@@ -64,6 +66,22 @@ export default function Sidebar({
...
```
saltyskip, line 79: `createLayout` (FLIP animation) records element positions before a DOM change and animates them to new positions. But tab switching here completely swaps DOM children — agents list vs flows list share no elements. The recorded elements get unmounted and the new elements didn't exist during `record()`, so there's nothing to animate between.

This will likely produce no visible animation or flicker. For a tab content crossfade/slide, consider a simpler CSS transition approach (e.g., opacity + transform on mount/unmount) or use `animate()` directly on the container. Alternatively, render both tabs and toggle `display: none` so `createLayout` can actually track persistent elements.

cthulu-studio/src/components/Sidebar.tsx
```diff
@@ -64,6 +66,22 @@ export default function Sidebar({
...
+  function switchTab(tab: "agents" | "flows") {
+    layoutRef.current?.record();
+    flushSync(() => setActiveTab(tab));
```
saltyskip, line 82: `flushSync` forces synchronous re-render, which is expensive and should be used sparingly. If the `createLayout` FLIP approach is replaced with CSS transitions, this can be removed entirely. If you keep the FLIP pattern, `flushSync` is necessary for it to work, but the FLIP approach itself is the wrong tool here (see other comment).

cthulu-studio/src/components/Sidebar.tsx
```diff
@@ -64,6 +66,22 @@ export default function Sidebar({
   const [prompts, setPrompts] = useState<SavedPrompt[]>([]);
   const [agentMeta, setAgentMeta] = useState<Map<string, { busy: boolean; sessions: InteractSessionInfo[]; cost: number }>>(new Map());
   const [expandedAgents, setExpandedAgents] = useState<Set<string>>(new Set());
+  const [activeTab, setActiveTab] = useState<"agents" | "flows">("agents");
+  const tabContentRef = useRef<HTMLDivElement>(null);
+  const layoutRef = useRef<ReturnType<typeof createLayout> | null>(null);
+
+  useEffect(() => {
```
saltyskip, line 73: If this component ever remounts (e.g., parent key change), `tabContentRef.current` could be null on the first render pass before React attaches the ref. The `if` guard handles that, but the layout would silently never be created.

Minor: consider a ref callback pattern instead if you want to guarantee setup:
```tsx
const tabContentRef = useCallback((node: HTMLDivElement | null) => {
  layoutRef.current?.revert();
  if (node) {
    layoutRef.current = createLayout(node, { duration: 300, ease: "out(3)" });
  }
}, []);
```

saltyskip (changes requested):
## 🤖 Cthulu Review

**The core animation approach won't work as intended.** `createLayout` (FLIP) is designed for animating persistent elements to new positions — but tab switching here completely unmounts one list and mounts another. The recorded elements disappear and the new elements weren't tracked, so the animation has nothing to interpolate between.

### Issues

- **FLIP misuse**: `createLayout` + `record()` / `animate()` requires shared DOM elements across state changes. Conditional rendering (`activeTab === "agents" ? ... : ...`) destroys and recreates elements, breaking this assumption.
- **`flushSync` overhead**: Only needed because of the FLIP pattern. If animation approach changes, this can be removed.

### Suggestions

1. **CSS transitions**: Simplest — fade or slide the tab content with `opacity`/`transform` transitions on class change.
2. **Render both tabs, toggle visibility**: Keep both lists mounted with `display: none` on the inactive one. This lets `createLayout` track elements, but adds complexity for no clear benefit over CSS.
3. **`animate()` directly**: Use animejs `animate()` on the container for a simple entrance animation on tab switch.

The tab bar UI and styling look good — this is just about the animation implementation.

---
_Automated review by Cthulu_

saltyskip (approved):

PR #114, opened 2026-03-19
feat: self-hosted authentication with JWT + per-user data isolation

MPRR-Pandu:
## Summary

- Self-hosted email/password auth with bcrypt hashing and JWT token issuance
- `AuthUser` Axum extractor with dev mode bypass (`AUTH_ENABLED=false` = no auth)
- Per-user directory helpers (`user_data_dir`, `user_key`, `ensure_user_dirs`)
- Login/signup React UI with auth gate in Studio
- Auth token injection on all API requests (Bearer header)
- Better Auth setup doc for future migration to managed auth

## Endpoints

| POST | `/api/auth/signup` | Create account (email + password 8+) |
| POST | `/api/auth/login` | Login, returns JWT |

## Config

| Env Var | Default | Description |
|---------|---------|-------------|
| `AUTH_ENABLED` | `false` | Enable auth (set to `true`) |
| `VITE_AUTH_ENABLED` | unset | Enable auth gate in Studio UI |

## Verification

- `cargo check` — 0 errors
- `cargo test` — 317 passed


saltyskip:

_Reviewing PR #114 — this may take a few minutes._

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,447 @@
...
+    match std::fs::read_to_string(&path) {
+        Ok(secret) if !secret.trim().is_empty() => secret.trim().to_string(),
+        _ => {
+            let secret = uuid::Uuid::new_v4().to_string();
+            let _ = std::fs::write(&path, &secret);
+            secret
+        }
+    }
+}
+
```
saltyskip, line 92: **Weak JWT secret**: `uuid::Uuid::new_v4()` produces 122 bits of entropy in a predictable hex-dash format. For HMAC-SHA256 signing you should use a full 256-bit random key. Consider:
```rust
use rand::Rng;
let secret: String = rand::rng().sample_iter(&rand::distr::Alphanumeric).take(64).collect();
```
A UUID is guessable if an attacker knows the approximate generation time + system state.

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,447 @@
...
+pub fn auth_enabled() -> bool {
+    std::env::var("AUTH_ENABLED")
+        .map(|v| v == "true")
+        .unwrap_or(false)
+}
+
+impl FromRequestParts<AppState> for AuthUser {
```
saltyskip, line 143: **`auth_enabled()` re-reads `std::env::var` on every request** instead of using the `config.auth_enabled` field you already added to `Config`. This means:
1. The `Config::auth_enabled` field is dead code
2. Every request does a syscall to read the env var
3. Auth state could change mid-flight if something mutates the env

Pass `auth_enabled` through `AppState` and use that instead of re-reading the env var.

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,447 @@
...
+                Json(json!({ "error": "User already exists" })),
+            );
+        }
+    }
+
+    // Hash password
+    let password_hash = match bcrypt::hash(&body.password, 12) {
+        Ok(h) => h,
+        Err(e) => {
+            return (
```
saltyskip, line 251: **TOCTOU race condition in signup**: The read lock is dropped (line 250-257) before the write lock is acquired (line 290). Two concurrent signups with the same email can both pass the existence check and both insert, with the second silently overwriting the first user's password hash.

Fix: hold a single write lock for the entire check-then-insert:
```rust
}
store.insert(user.clone());
```

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,447 @@
...
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(json!({ "error": format!("Hash failed: {e}") })),
+            );
```
saltyskip, line 254: **`bcrypt::hash` with cost 12 blocks the tokio runtime**. bcrypt is CPU-intensive (~250ms at cost 12) and this runs on an async task, starving other requests. Wrap in `spawn_blocking`:
```rust
let password_hash = tokio::task::spawn_blocking(move || bcrypt::hash(&password, 12))
    .await
    .map_err(|e| /* ... */)??
```
Same issue in `login` with `bcrypt::verify` (line 323).

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,447 @@
...
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("users.json");
+        let json = serde_json::to_string_pretty(&self.users)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn find_by_email(&self, email: &str) -> Option<&StoredUser> {
+        self.users.get(email)
+    }
+
```
saltyskip, line 60: **Non-atomic file write** violates project convention. CLAUDE.md rule #9 says "Always write via temp file + rename". A crash during `std::fs::write` will corrupt `users.json`. Use:
```rust
let tmp_path = path.with_extension("json.tmp");
std::fs::write(&tmp_path, &json)?;
std::fs::rename(&tmp_path, &path)?;
```

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,447 @@
+//! Self-hosted authentication: signup, login, JWT issuance + verification.
```
saltyskip, line 1: **File naming**: This file is named `clerk_auth.rs` but implements self-hosted bcrypt/JWT auth — Clerk is not involved. The commit history shows a pivot from Clerk to self-hosted, but the naming wasn't updated. Should be renamed to something like `auth.rs` to avoid confusion (the `api/mod.rs` already has an `auth` module — may need to reconcile).

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,447 @@
...
+    {
+        let mut store = state.user_store.write().await;
+        store.insert(user.clone());
+        let _ = store.save(&state.data_dir);
+    }
+
+    (
+        StatusCode::CREATED,
+        Json(json!({
+            "token": token,
```
saltyskip, line 299: Silent error swallowing: `let _ = store.save(&state.data_dir);` — if the save fails, the user gets a success response with a JWT token, but their account isn't persisted. On server restart they won't be able to log in. At minimum log the error; ideally return 500.

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,114 @@
...
+  if (!authEnabled) return <>{children}</>;
+  if (token) {
+    // Set token getter so all API requests include it
+    setAuthTokenGetter(() => Promise.resolve(token));
+    return <>{children}</>;
+  }
+
+  return <AuthForm onSuccess={(t) => {
+    localStorage.setItem("cthulu_auth_token", t);
+    setToken(t);
```
saltyskip, line 32: **`setAuthTokenGetter` called on every render**: When `token` is truthy, this component re-runs `setAuthTokenGetter(() => Promise.resolve(token))` on every render cycle, creating a new closure each time. Wrap in a `useEffect` or move the token getter setup into the `onSuccess` callback only (which you already do on line 35). The duplicate here is unnecessary.

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,114 @@
...
+export default function AuthGate({ children }: AuthGateProps) {
+  const authEnabled = import.meta.env.VITE_AUTH_ENABLED === "true";
+  const [token, setToken] = useState<string | null>(
+    () => localStorage.getItem("cthulu_auth_token"),
+  );
+
+  if (!authEnabled) return <>{children}</>;
+  if (token) {
+    // Set token getter so all API requests include it
+    setAuthTokenGetter(() => Promise.resolve(token));
```
saltyskip: **No way to log out or handle expired tokens**: If the JWT expires (24h), the user has a stale token in localStorage that will cause 401s on every request with no recovery path. Consider:
1. Adding a logout button that clears localStorage + token state
2. Intercepting 401 responses in `apiFetch` to clear the token and redirect to login

cthulu-backend/api/user_context.rs
```diff
@@ -0,0 +1,70 @@
...
+/// Dev mode: returns `state.data_dir` directly.
+/// Auth mode: returns `state.data_dir/users/{user_id}`.
+pub fn user_data_dir(state: &AppState, auth: &AuthUser) -> PathBuf {
+    if auth_enabled() {
+        state.data_dir.join("users").join(&auth.user_id)
+    } else {
+        state.data_dir.clone()
+    }
+}
+
```
saltyskip, line 22: Same issue as `clerk_auth.rs`: `auth_enabled()` reads the env var on every call instead of using config state. This function is called on every request that needs user context — should take `auth_enabled: bool` from `AppState` instead.

saltyskip (changes requested):
## 🤖 Cthulu Review

Good foundation for self-hosted auth. The overall structure (AuthUser extractor, dev mode bypass, per-user directories, frontend auth gate) is solid. A few issues need fixing before merge:

### Blocking

1. **TOCTOU race in signup** — read lock released before write lock acquired; concurrent signups with the same email can overwrite each other. Hold a single write lock for check-then-insert.

2. **bcrypt blocks tokio runtime** — `bcrypt::hash` (cost 12, ~250ms) and `bcrypt::verify` run on async tasks, starving the thread pool. Wrap in `spawn_blocking`.

3. **`auth_enabled()` re-reads env var on every request** — the `Config::auth_enabled` field you added is dead code. Pass it through `AppState` and use it consistently in the extractor and `user_context.rs`.

4. **Non-atomic `users.json` write** — violates project rule #9. Use temp file + rename pattern.

5. **Weak JWT secret** — UUID v4 has predictable structure. Use a proper 256-bit random key.

6. **Silent save failure** — signup returns success + JWT even if `store.save()` fails, leaving the user unable to log in after restart.

### Non-blocking

- File is named `clerk_auth.rs` but contains self-hosted auth — rename to avoid confusion
- No logout mechanism or expired-token handling in the frontend
- `setAuthTokenGetter` called redundantly on every render when token exists

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #114 (previous HEAD: `f91b02e`, new HEAD: `17a1268`)_

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,458 @@
...
+    match std::fs::read_to_string(&path) {
+        Ok(secret) if !secret.trim().is_empty() => secret.trim().to_string(),
+        _ => {
+            let secret = uuid::Uuid::new_v4().to_string();
+            std::fs::write(&path, &secret)
+                .unwrap_or_else(|e| panic!("Failed to write JWT secret to {}: {e}", path.display()));
+            secret
+        }
+    }
+}
```
saltyskip, line 91: **Security: Weak JWT secret.** `Uuid::new_v4()` provides ~122 bits of entropy and is not designed as a cryptographic key. Use a proper CSPRNG instead:

```rust
use rand::Rng;
let secret: String = rand::thread_rng()
    .gen::<[u8; 32]>()
    .iter()
    .map(|b| format!("{b:02x}"))
    .collect();
```

Or use `base64` (which is already added to `Cargo.toml` but unused) to encode 32 random bytes.

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,458 @@
+//! Self-hosted authentication: signup, login, JWT issuance + verification.
```
saltyskip, line 1: **Naming: Module is called `clerk_auth` but contains zero Clerk integration.** This is self-hosted bcrypt/JWT auth. The name will confuse future contributors. Rename to `auth.rs` or `self_auth.rs`. (I see there's already an `auth` module — consider merging or picking a clearer name like `local_auth`.)

Cargo.toml
```diff
@@ -22,6 +22,9 @@ futures = "0.3.32"
 async-stream = "0.3.6"
 serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.149"
+jsonwebtoken = "9"
+base64 = "0.22"
+bcrypt = "0.19"
```
saltyskip, line 27: **Unused dependency.** `base64` is added here but never imported or used anywhere in the changed code. Either use it (e.g., for encoding the JWT secret) or remove it to avoid bloat.

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,117 @@
...
+  }, [token]);
+
+  if (!authEnabled) return <>{children}</>;
+  if (token) return <>{children}</>;
+
+  return <AuthForm onSuccess={(t) => {
+    localStorage.setItem("cthulu_auth_token", t);
+    setToken(t);
+  }} />;
+}
```
saltyskip, line 34: **Bug: Expired tokens leave the UI broken.** When `token` is non-null (from localStorage) but expired, `AuthGate` renders `children` — but all API calls will 401. The user sees a broken app with no way to re-authenticate.

Two fixes needed:
1. Add a 401 handler in `apiFetch` that clears the token and forces re-render to login:
```ts
if (res.status === 401) {
  localStorage.removeItem('cthulu_auth_token');
  window.location.reload();
}
```
2. Or validate the token client-side before rendering children (decode the JWT and check `exp`).

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+function AuthForm({ onSuccess }: { onSuccess: (token: string) => void }) {
+  const [mode, setMode] = useState<"login" | "signup">("login");
+  const [email, setEmail] = useState("");
+  const [password, setPassword] = useState("");
+  const [error, setError] = useState("");
+  const [loading, setLoading] = useState(false);
+
```
saltyskip, line 42: **Missing: No logout UI.** Once authenticated, the user has no way to sign out. Consider adding a logout handler (clear localStorage, reset token state) and expose it — even if the logout button is wired up in a follow-up PR, the handler should exist here.

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,458 @@
...
+
+    if let Some(query) = parts.uri.query() {
+        for pair in query.split('&') {
+            if let Some(token) = pair.strip_prefix("token=") {
+                return Some(token.to_string());
+            }
+        }
+    }
+
+    None
```
saltyskip, line 204: **Security note:** Accepting tokens via `?token=` query parameter means JWTs will appear in server access logs, browser history, and `Referer` headers. This is fine for SSE EventSource (which can't set headers), but consider documenting that this path is only intended for SSE and not general API use.

cthulu-backend/api/user_context.rs
```diff
@@ -0,0 +1,70 @@
...
+/// Returns the per-user data directory.
+/// Dev mode: returns `state.data_dir` directly.
+/// Auth mode: returns `state.data_dir/users/{user_id}`.
+pub fn user_data_dir(state: &AppState, auth: &AuthUser) -> PathBuf {
+    if state.auth_enabled {
+        state.data_dir.join("users").join(&auth.user_id)
+    } else {
+        state.data_dir.clone()
+    }
+}
```
saltyskip, line 21: **Dead code.** `user_data_dir`, `user_key`, and `ensure_user_dirs` are defined but not called anywhere in this PR. This will trigger `dead_code` warnings unless suppressed. If this is scaffolding for a follow-up PR, add `#[allow(dead_code)]` or defer to the PR that actually uses these.

saltyskip (changes requested):
## 🤖 Cthulu Review

Good foundation for self-hosted auth. The JWT/bcrypt flow is solid, the `AuthUser` extractor pattern is clean, and the dev-mode bypass is well thought out. However, there are two blocking issues before merge:

### Blocking

1. **Weak JWT secret** — `Uuid::new_v4()` is not a cryptographic key generator. Use 32 bytes from a CSPRNG (e.g., `rand::thread_rng().gen::<[u8; 32]>()`). This is a security-critical path.

2. **Expired token = broken UI** — When a JWT in localStorage expires, `AuthGate` still renders the app (token is non-null), but all API calls will 401. The user has no way to recover without manually clearing localStorage. Need a 401 handler in `apiFetch` that clears the token and reloads.

### Non-blocking

- **Module naming**: `clerk_auth.rs` contains zero Clerk code — rename to `local_auth.rs` or similar to avoid confusion.
- **Unused `base64` dep** in `Cargo.toml` — remove or use it.
- **`user_context.rs` is dead code** — will produce compiler warnings. Add `#[allow(dead_code)]` or defer to the PR that uses it.
- **No logout UI** — users can't sign out. Even if deferred, the mechanism should be planned.
- **No rate limiting** on `/auth/login` — brute force is trivial. Fine for v1 but track it.

---

_Automated review by Cthulu_

MPRR-Pandu:
#117 (A) + #118    │ Backend deps/config → A, auth module → B,    │ (B) + #119 (C)

Covered in 117 , 118 , 119

PR #115, opened 2026-03-19
feat: user profiles + teams with member management

MPRR-Pandu:
## Summary

Depends on: #12 (auth foundation)

- Add `name` + `avatar_url` fields to user profiles (GET/PUT `/api/auth/me`)
- Teams: flat groups of users stored in `teams.json` (no roles)
- Team CRUD + member management by email
- User search endpoint (search by name or email, 300ms debounced)
- Profile dropdown in TopBar with expandable team cards
- Search-as-you-type member picker with inline results

## Endpoints

| GET | `/api/auth/me` | Get user profile |
| PUT | `/api/auth/me` | Update name/avatar |
| GET | `/api/auth/users/search?q=` | Search users by name/email |
| GET | `/api/teams` | List user's teams |
| POST | `/api/teams` | Create team |
| GET | `/api/teams/{id}` | Get team with members |
| POST | `/api/teams/{id}/members` | Add member by email |
| DELETE | `/api/teams/{id}/members/{uid}` | Remove member |

## Verification

- `cargo check` — 0 errors
- `cargo test` — 319 passed


saltyskip:

_Reviewing PR #115 — this may take a few minutes._

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,545 @@
...
+    match std::fs::read_to_string(&path) {
+        Ok(secret) if !secret.trim().is_empty() => secret.trim().to_string(),
+        _ => {
+            let secret = uuid::Uuid::new_v4().to_string();
+            let _ = std::fs::write(&path, &secret);
+            secret
+        }
+    }
+}
+
```
saltyskip, line 90: **Security: Weak JWT secret** — `uuid::Uuid::new_v4()` produces 122 bits of entropy, but it's formatted as a hex string and used as an HMAC key via `from_secret(secret.as_bytes())`. The effective key is the *string bytes*, not raw random bytes. More importantly, if the write to disk fails (line 91 silently discards the error with `let _ =`), the secret is lost on restart and all existing JWTs are silently invalidated.

Consider:
1. Use a proper CSPRNG (e.g., `rand::thread_rng().gen::<[u8; 32]>()` then hex-encode for storage)
2. Don't silently ignore the write failure — if the secret can't be persisted, the server should probably refuse to start

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,545 @@
...
+    {
+        let mut store = state.user_store.write().await;
+        store.insert(user.clone());
+        let _ = store.save(&state.data_dir);
+    }
+
+    (
+        StatusCode::CREATED,
+        Json(json!({
+            "token": token,
```
saltyskip, line 298: **TOCTOU race condition on signup** — The existence check (line 254, read lock) and the insert (line 296, write lock) are in separate critical sections. Two concurrent signups with the same email can both pass the check and both insert, with the second silently overwriting the first user (different `id`, different `password_hash`). This effectively hijacks the first user's account.

Fix: hold a single write lock for the entire check-and-insert:
```rust
}
store.insert(user.clone());
```
The bcrypt hash can be computed before acquiring the lock.

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,545 @@
...
+        Self { users }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("users.json");
+        let json = serde_json::to_string_pretty(&self.users)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn find_by_email(&self, email: &str) -> Option<&StoredUser> {
```
saltyskip, line 58: **Data loss: Non-atomic write** — `std::fs::write(path, json)` is not atomic. A crash mid-write corrupts `users.json`. The project's own CLAUDE.md (rule 9) mandates atomic writes via temp-file + rename. Apply the same pattern here:
```rust
std::fs::write(&tmp, json)?;
std::fs::rename(tmp, path)?;
```
Same issue in `TeamStore::save`.

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    // Hash password
+    let password_hash = match bcrypt::hash(&body.password, 12) {
+        Ok(h) => h,
+        Err(e) => {
+            return (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(json!({ "error": format!("Hash failed: {e}") })),
+            );
```
saltyskip, line 264: **Performance: bcrypt cost 12 blocks the Tokio runtime** — `bcrypt::hash` with cost 12 takes ~250ms and is CPU-bound. Since this runs on an async task, it blocks the Tokio worker thread. Use `tokio::task::spawn_blocking` to move it off the async runtime:
```rust
let password_hash = tokio::task::spawn_blocking(move || bcrypt::hash(&password, 12))
    .await
    .map_err(|e| /* ... */)?
    .map_err(|e| /* ... */)?;
```
Same issue on line 329 with `bcrypt::verify`.

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,545 @@
...
+    drop(store);
+
+    // Save outside the mutable borrow
+    let store = state.user_store.read().await;
+    let _ = store.save(&state.data_dir);
+
+    (StatusCode::OK, Json(json!({
+        "id": updated.id,
+        "email": updated.email,
+        "name": updated.name,
```
saltyskip, line 411: **Bug: save after drop doesn't persist the update** — Line 408 drops the write guard, then line 411 acquires a *read* lock and calls `save()`. This works only if no other write happens between the drop and the re-acquire. But the real issue is that `save()` takes `&self` (immutable), yet it should conceptually be called while you still hold the write lock to guarantee the data written matches what was just modified.

Simpler fix: call `store.save()` before dropping the write guard:
```rust
let updated = user.clone();
drop(store);
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    /// List all teams a user belongs to.
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
```
saltyskip, line 54: Same non-atomic write issue as `UserStore::save`. Use temp-file + rename per project rule 9.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,249 @@
...
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
+            // If no members left, delete the team
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
```
saltyskip, line 198: **Bug: any member can remove any other member** — There's no authorization check beyond "is the caller a member of this team". Any team member can remove any other member, including the creator. Consider restricting removal to: (a) the team creator, or (b) self-removal only. At minimum, prevent removing the last admin/creator.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,249 @@
...
+    match store.teams.get(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            // Resolve member emails for display
+            let user_store = state.user_store.read().await;
+            let members: Vec<Value> = team.members.iter().map(|uid| {
+                let user = user_store.users.values().find(|u| u.id == *uid);
+                json!({
+                    "id": uid,
+                    "email": user.map(|u| u.email.as_str()).unwrap_or("unknown"),
+                    "name": user.and_then(|u| u.name.as_deref()),
```
saltyskip, line 128: **Deadlock risk: nested read locks** — `get_team` holds a read lock on `team_store` (line 124), then acquires a read lock on `user_store` (line 128). If another handler acquires these locks in the opposite order, this will deadlock. Always acquire locks in a consistent order across all handlers, or clone the data you need and drop the first lock before acquiring the second.

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,152 @@
...
+  if (!authEnabled) return <>{children}</>;
+
+  if (token) {
+    setAuthTokenGetter(() => Promise.resolve(token));
+    return (
+      <AuthContext.Provider value={{ logout, userEmail: emailFromToken(token) }}>
+        {children}
+      </AuthContext.Provider>
+    );
+  }
```
saltyskip, line 62: **Bug: `setAuthTokenGetter` called during render** — Line 62 calls `setAuthTokenGetter()` unconditionally during the render phase (not inside a useEffect or callback). This is a side effect during render, which React may execute multiple times in StrictMode or concurrent features. It also means `setAuthTokenGetter` is called on every re-render.

Wrap this in a `useEffect`:
```tsx
useEffect(() => {
}, [token]);
```

.superpowers/brainstorm/77895-1773893838/.server-info
```diff
@@ -0,0 +1 @@
+{"type":"server-started","port":64600,"host":"127.0.0.1","url_host":"localhost","url":"http://localhost:64600","screen_dir":"/Users/mundlapatipandurangaraju/Desktop/zanoWallets/dev/_newclone/cthulu/.superpowers/brainstorm/77895-1773893838"}
```
saltyskip, line 1: **Accidental commit: `.superpowers/` directory** — This contains local dev tool state (server PID, log files, brainstorm HTML artifacts) and another developer's filesystem paths. Should be `.gitignore`'d and removed from this PR. Also applies to `.server.log`, `.server.pid`, `theme-dropdown-options.html`, and `waiting.html`.

MEMORY.md
```diff
@@ -0,0 +1,223 @@
+# Cthulu -- Session Memory
```
saltyskip, line 1: **Should not be committed** — This is a Claude Code session memory file (`MEMORY.md`). It contains AI assistant context, not project documentation. It should be in `.gitignore`, not checked into the repo.

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,545 @@
...
+) -> (StatusCode, Json<Value>) {
+    let store = state.user_store.read().await;
+    // Find user by ID (scan values since store is keyed by email)
+    let user = store.users.values().find(|u| u.id == auth.user_id);
+    match user {
+        Some(u) => (StatusCode::OK, Json(json!({
+            "id": u.id,
+            "email": u.email,
+            "name": u.name,
+            "avatar_url": u.avatar_url,
```
saltyskip, line 365: **O(n) user lookup by ID** — `get_profile` and `update_profile` both scan all users to find one by ID (the store is keyed by email). This is O(n) per request. Consider adding a secondary index (`HashMap<String, String>` mapping `id -> email`) or re-keying the store by ID with an email index instead.

saltyskip (changes requested):
## 🤖 Cthulu Review

### Blocking Issues

1. **Signup TOCTOU race condition** — Two concurrent signups with the same email can both succeed, with the second silently overwriting the first user's account (different ID and password hash). This is an account hijack vulnerability.

2. **Non-atomic file writes** — Both `UserStore::save` and `TeamStore::save` use `std::fs::write` directly, violating the project's own rule 9 (atomic YAML persistence via temp-file + rename). A crash mid-write corrupts `users.json` or `teams.json`.

3. **bcrypt blocks Tokio runtime** — `bcrypt::hash` (cost 12, ~250ms) and `bcrypt::verify` run synchronously on async worker threads, blocking the entire Tokio runtime during auth operations. Must use `spawn_blocking`.

4. **Accidental files committed** — `.superpowers/` directory (local dev server state, PID files, another developer's filesystem paths) and `MEMORY.md` (Claude Code session memory) should not be in the repo. Add to `.gitignore` and remove from the PR.

5. **Side effect during React render** — `AuthGate.tsx` calls `setAuthTokenGetter()` directly in the render body, which React may call multiple times. Needs to be in a `useEffect`.

### Non-blocking Concerns

- JWT secret generation uses UUID v4 string bytes rather than proper CSPRNG output, and silently ignores write failures
- O(n) user-by-ID lookups (store keyed by email, but profile/update scan by ID)
- Any team member can remove any other member including the creator — no authorization hierarchy
- Potential deadlock from nested read locks acquired in inconsistent order (team_store → user_store)
- `update_profile` drops the write lock then re-acquires a read lock to save — fragile ordering

### PR Size

The PR description states a ~2 file / ~100 line per-PR discipline, but this PR touches 31 files with 8500+ lines added. The actual feature code is well-structured, but the bundled docs/plans/specs and accidental files inflate it significantly. Consider splitting the docs into a separate PR.

---

_Automated review by Cthulu_

MPRR-Pandu:
#120 (D) + #121 (E) covered in 120 and 121 PRs

PR #116, opened 2026-03-19
feat: team sessions — shared agents with multi-user broadcast

MPRR-Pandu:
## Summary

Depends on: #13 (profiles + teams)

- Add `team_id` field to Agent struct — when set, agent belongs to a team
- Team membership check on chat endpoint (403 for non-members)
- Broadcast `user_message` SSE event with sender attribution so all subscribers see who asked
- Frontend handles `user_message` event in chat stream
- Sidebar splits agents into "Agents" (personal) and per-team sections
- Visual distinction: accent border + "team" badge on team agents
- Agent create picker: "+" dropdown offers Personal Agent or Team Agent (per team)
- Serial turn-taking: one sender at a time, others spectate via existing reconnect endpoint

## How It Works

1. Create a team agent via the "+" picker (select a team)
2. All team members see it in their team's sidebar section
3. Any member sends a message — `user_message` event broadcasts to all subscribers
4. Other members join via existing reconnect endpoint (GET `/agents/{id}/sessions/{sid}/chat/stream`)
5. One sender at a time (409 while busy), everyone spectates in real-time

## Security

- Slack bot tokens never leave the user's machine (per CLAUDE.md security rule)
- Team agents cannot run Slack-based flows as shared sessions

## Verification

- `cargo check` — 0 errors
- `cargo test` — 319 passed


saltyskip:

_Reviewing PR #116 — this may take a few minutes._

MPRR-Pandu (commented):
## Review

PR #116 builds on PRs #114 and #115. It inherits all blocking issues from those PRs and adds its own. The team agent + broadcast feature is well-structured, but the inherited issues must be fixed before merge.

### Blocking Issues (inherited from #114/#115, still present here)

1. **TOCTOU race condition in signup** — `clerk_auth.rs` read lock is dropped before the write lock is acquired. Two concurrent signups with the same email can both succeed, with the second overwriting the first user's account. Hold a single write lock for the entire check-then-insert operation.

2. **bcrypt blocks Tokio runtime** — `bcrypt::hash` (cost 12, ~250ms) and `bcrypt::verify` run synchronously on async worker threads. Wrap both in `tokio::task::spawn_blocking`.

3. **Non-atomic file writes** — Both `UserStore::save` and `TeamStore::save` use `std::fs::write` directly, violating project rule 9 (temp-file + rename). A crash mid-write corrupts the data files.

4. **Accidental files committed** — `.superpowers/` directory (PID files, server logs, dev filesystem paths) and `MEMORY.md` (Claude Code session memory) should not be in the repo. Add to `.gitignore` and remove.

5. **Side effect during React render** — `AuthGate.tsx` calls `setAuthTokenGetter()` directly in the render body. This should be in a `useEffect`.

6. **`auth_enabled()` re-reads env var on every request** — The `Config::auth_enabled` field is dead code. Pass it through `AppState` and use it consistently.

### Blocking Issues (new in this PR)

7. **`user_message` broadcast format** — The `user_message` SSE event (`chat.rs:1455-1461`) uses `format!("user_message:{}", serde_json::to_string(...))` which is a custom format rather than a proper SSE `data:` field. The client-side `useAgentChat.ts` needs to parse this format correctly. Verify the frontend event parsing handles this custom prefix format.

### Non-blocking

- JWT secret generation uses UUID v4 rather than proper CSPRNG — acceptable for dev/self-hosted but not production
- O(n) user-by-ID lookups in profile endpoints (store keyed by email, profile lookups scan by ID)
- Any team member can remove any other member including the creator — no authorization hierarchy
- File named `clerk_auth.rs` but implements self-hosted auth — should rename
- No logout mechanism or expired-token handling in the frontend (24h JWT with no refresh)
- PR size far exceeds the project's stated ~2 files / ~100 lines per PR discipline (35 files, 10k+ lines)
- The bundled docs, plans, specs, and brainstorm artifacts inflate the PR significantly — split docs into a separate PR

---

_Review by code-reviewer_

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,545 @@
...
+        if store.find_by_email(&email).is_some() {
+            return (
+                StatusCode::CONFLICT,
+                Json(json!({ "error": "User already exists" })),
+            );
+        }
+    }
+
+    // Hash password
+    let password_hash = match bcrypt::hash(&body.password, 12) {
```
saltyskip, line 258: **Bug: `bcrypt::hash` blocks the async runtime.** bcrypt cost 12 takes ~250ms on a single core. This runs inside an async handler without `spawn_blocking`, stalling the entire Tokio worker thread and all concurrent requests.

```rust
let password_hash = tokio::task::spawn_blocking(move || {
    bcrypt::hash(&body.password, 12)
}).await.map_err(|e| /* ... */)?;
```

Same issue in the `login` handler with `bcrypt::verify`.

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,545 @@
...
+
+const JWT_EXPIRY_HOURS: i64 = 24;
+
+/// Load or generate the JWT signing secret.
+pub fn load_or_create_jwt_secret(data_dir: &PathBuf) -> String {
+    let path = data_dir.join("jwt_secret");
+    match std::fs::read_to_string(&path) {
+        Ok(secret) if !secret.trim().is_empty() => secret.trim().to_string(),
+        _ => {
+            let secret = uuid::Uuid::new_v4().to_string();
```
saltyskip, line 84: **Weak JWT secret.** UUID v4 provides ~122 bits of randomness, which is below the recommended 256 bits for HMAC-SHA256 signing keys. Consider generating a 32-byte random secret instead:

```rust
use rand::RngCore;
let mut bytes = [0u8; 32];
rand::thread_rng().fill_bytes(&mut bytes);
let secret = hex::encode(bytes);
```

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,545 @@
...
+        user.name = Some(name);
+    }
+    if let Some(avatar_url) = body.avatar_url {
+        user.avatar_url = Some(avatar_url);
+    }
+    let updated = user.clone();
+    drop(store);
+
+    // Save outside the mutable borrow
+    let store = state.user_store.read().await;
```
saltyskip, line 405: **Data race in `update_profile`.** You drop the write lock on line 402, then re-acquire a read lock to call `save()`. Between those two operations, another concurrent request could acquire a write lock and modify the store — then your `save()` persists a partially-updated state. Keep the write lock for the save:

```rust
// ... mutate user ...
let updated = user.clone();
drop(store);
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,249 @@
...
+
+async fn remove_member(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path((id, user_id)): Path<(String, String)>,
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
```
saltyskip, line 192: **Any team member can remove any other member, including the team creator.** There's no authorization distinction — member A can kick member B (or the creator). At minimum, restrict removal to: (1) the team creator can remove anyone, (2) a member can only remove themselves.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub fn load(data_dir: &PathBuf) -> Self {
+        let path = data_dir.join("teams.json");
+        let teams = match std::fs::read_to_string(&path) {
+            Ok(contents) => serde_json::from_str(&contents).unwrap_or_default(),
+            Err(_) => HashMap::new(),
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
```
saltyskip, line 47: **Non-atomic file write.** CLAUDE.md rule #9 requires temp-file + rename for persistence. `std::fs::write` can leave a truncated file on crash. Use `path.with_extension("json.tmp")` then `std::fs::rename`. Same issue in `UserStore::save`.

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const logout = useCallback(() => {
+    localStorage.removeItem("cthulu_auth_token");
+    setAuthTokenGetter(null);
+    setToken(null);
+  }, []);
+
+  if (!authEnabled) return <>{children}</>;
+
+  if (token) {
+    setAuthTokenGetter(() => Promise.resolve(token));
```
saltyskip, line 56: **Side effect during render.** `setAuthTokenGetter(...)` is called during the render phase (not inside a `useEffect` or event handler). Renders should be pure in React. Move this into a `useEffect`:

```tsx
useEffect(() => {
}, [token]);
```

.superpowers/brainstorm/77895-1773893838/.server-info
```diff
@@ -0,0 +1 @@
+{"type":"server-started","port":64600,"host":"127.0.0.1","url_host":"localhost","url":"http://localhost:64600","screen_dir":"/Users/mundlapatipandurangaraju/Desktop/zanoWallets/dev/_newclone/cthulu/.superpowers/brainstorm/77895-1773893838"}
```
saltyskip, line 1: **Accidental commit.** The `.superpowers/brainstorm/` directory contains local dev artifacts from another contributor's machine (path shows `/Users/mundlapatipandurangaraju/`). These should be removed from the PR and `.superpowers/` added to `.gitignore`.

MEMORY.md
```diff
@@ -0,0 +1,223 @@
+# Cthulu -- Session Memory
```
saltyskip, line 1: **Unrelated to this PR.** This is a 223-line AI session memory file. It's not related to team sessions and will go stale quickly. Should be in a separate PR or excluded.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Requesting changes** — there are a few issues that should be addressed before merging:

### Blocking

1. **bcrypt blocks the async runtime** — `bcrypt::hash` (cost 12, ~250ms) and `bcrypt::verify` run synchronously inside async handlers. This will stall the Tokio worker thread and degrade all concurrent request handling. Wrap in `tokio::task::spawn_blocking`.

2. **Data race in `update_profile`** — The write lock is dropped before `save()`, allowing a concurrent writer to interleave. Keep the write lock across the mutation + save.

3. **Any team member can remove any other member** — No authorization check in `remove_member`. A regular member can kick the team creator. Add role-based or creator-only removal logic.

4. **Side effect during render in `AuthGate`** — `setAuthTokenGetter()` is called during the React render phase, violating React's purity rules. Move to a `useEffect`.

5. **Accidental `.superpowers/` directory committed** — Contains local dev server artifacts from another contributor's machine. Remove and add to `.gitignore`.

6. **`MEMORY.md` (223 lines) is unrelated** — AI session memory file bundled into a team-sessions PR. Should be a separate PR.

### Non-blocking

- **Non-atomic file persistence** — Both `UserStore::save` and `TeamStore::save` use `std::fs::write` directly instead of temp-file + rename per CLAUDE.md rule #9.
- **JWT secret entropy** — UUID v4 (~122 bits) is below the recommended 256 bits for HMAC-SHA256. Consider a 32-byte random secret.
- **Hooks violation in `ProfileMenu`** — Early return `if (!userEmail) return null` is placed before `useEffect` hooks, which will cause React to error on conditional hook calls.

### Docs/plans in the diff

~5,900 of the ~8,600 added lines are planning docs (`docs/superpowers/plans/`, `docs/superpowers/specs/`). If these are meant to be committed, they should be in a separate PR to keep this one reviewable.

---

_Automated review by Cthulu_

MPRR-Pandu:
covered in #122 (F)

PR #120, opened 2026-03-20
feat: teams backend with CRUD and member management

MPRR-Pandu:
## Summary

Depends on: #118 (auth module) — merge that first.

- Teams: flat groups of users, no roles — all members equal
- TeamStore with JSON persistence (`teams.json`)
- Team CRUD: list, create, get (with member email resolution)
- Member management: add by email, remove, auto-delete empty teams

## Endpoints

| GET | `/api/teams` | List user's teams |
| POST | `/api/teams` | Create team |
| GET | `/api/teams/{id}` | Get team with members |
| POST | `/api/teams/{id}/members` | Add member by email |
| DELETE | `/api/teams/{id}/members/{uid}` | Remove member |

## Files changed (4 files, ~250 lines)
- `teams.rs` — TeamStore, CRUD handlers, member management
- `api/mod.rs` — add teams module + team_store to AppState
- `routes.rs` — merge teams router
- `main.rs` — init team_store

## Verification
- `cargo check` — 0 errors
- `cargo test teams` — 2 passed

## Dependency chain
**PR D (4/6)** — merge after #118 (independent of PR C)

```
main → #117 → #118 → THIS PR D (teams backend)
                   └─ #119 PR C (frontend auth)
```

### DIFF
--- a/cthulu-backend/api/teams.rs
+++ b/cthulu-backend/api/teams.rs
@@ -1,0 +1,240 @@
+//! Teams: flat groups of users. No roles — all members are equal.
+//! Stored in `{data_dir}/teams.json`.
+
+use axum::extract::{Path, State};
+use axum::routing::{delete, get, post};
+use axum::{Json, Router};
+use hyper::StatusCode;
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Value};
+use std::collections::HashMap;
+use std::path::PathBuf;
+
+use crate::api::local_auth::AuthUser;
+use crate::api::AppState;
+
+// ── Team Model ───────────────────────────────────────────────
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Team {
+    pub id: String,
+    pub name: String,
+    pub created_by: String,
+    pub members: Vec<String>,
+    pub created_at: String,
+}
+
+/// In-memory team store backed by `{data_dir}/teams.json`.
+pub struct TeamStore {
+    pub teams: HashMap<String, Team>,
+}
+
+impl TeamStore {
+    pub fn load(data_dir: &PathBuf) -> Self {
+        let path = data_dir.join("teams.json");
+        let teams = match std::fs::read_to_string(&path) {
+            Ok(contents) => serde_json::from_str(&contents).unwrap_or_default(),
+            Err(_) => HashMap::new(),
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
+        self.teams.values()
+            .filter(|t| t.members.contains(&user_id.to_string()))
+            .collect()
+    }
+}
+
+// ── Routes ───────────────────────────────────────────────────
+
+pub fn router() -> Router<AppState> {
+    Router::new()
+        .route("/teams", get(list_teams).post(create_team))
+        .route("/teams/{id}", get(get_team))
+        .route("/teams/{id}/members", post(add_member))
+        .route("/teams/{id}/members/{user_id}", delete(remove_member))
+}
+
+// ── Handlers ─────────────────────────────────────────────────
+
+async fn list_teams(
+    auth: AuthUser,
+    State(state): State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.team_store.read().await;
+    let teams: Vec<&Team> = store.teams_for_user(&auth.user_id);
+    (StatusCode::OK, Json(json!({ "teams": teams })))
+}
+
+#[derive(Deserialize)]
+struct CreateTeamRequest {
+    name: String,
+}
+
+async fn create_team(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Json(body): Json<CreateTeamRequest>,
+) -> (StatusCode, Json<Value>) {
+    let name = body.name.trim().to_string();
+    if name.is_empty() {
+        return (StatusCode::BAD_REQUEST, Json(json!({ "error": "Team name required" })));
+    }
+
+    let team = Team {
+        id: uuid::Uuid::new_v4().to_string().replace('-', "_"),
+        name,
+        created_by: auth.user_id.clone(),
+        members: vec![auth.user_id.clone()],
+        created_at: chrono::Utc::now().to_rfc3339(),
+    };
+
+    let mut store = state.team_store.write().await;
+    let response = json!({ "team": &team });
+    store.teams.insert(team.id.clone(), team);
+    let _ = store.save(&state.data_dir);
+
+    (StatusCode::CREATED, Json(response))
+}
+
+async fn get_team(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path(id): Path<String>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.team_store.read().await;
+    match store.teams.get(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            let user_store = state.user_store.read().await;
+            let members: Vec<Value> = team.members.iter().map(|uid| {
+                let user = user_store.users.values().find(|u| u.id == *uid);
+                json!({
+                    "id": uid,
+                    "email": user.map(|u| u.email.as_str()).unwrap_or("unknown"),
+                    "name": user.and_then(|u| u.name.as_deref()),
+                })
+            }).collect();
+
+            (StatusCode::OK, Json(json!({
+                "team": {
+                    "id": team.id,
+                    "name": team.name,
+                    "created_by": team.created_by,
+                    "members": members,
+                    "created_at": team.created_at,
+                }
+            })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
+#[derive(Deserialize)]
+struct AddMemberRequest {
+    email: String,
+}
+
+async fn add_member(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path(id): Path<String>,
+    Json(body): Json<AddMemberRequest>,
+) -> (StatusCode, Json<Value>) {
+    let email = body.email.trim().to_lowercase();
+
+    let user_store = state.user_store.read().await;
+    let target_user = user_store.find_by_email(&email);
+    let target_id = match target_user {
+        Some(u) => u.id.clone(),
+        None => return (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
+    };
+    drop(user_store);
+
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            if team.members.contains(&target_id) {
+                return (StatusCode::CONFLICT, Json(json!({ "error": "Already a member" })));
+            }
+            team.members.push(target_id);
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
+async fn remove_member(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path((id, user_id)): Path<(String, String)>,
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
+// ── Tests ────────────────────────────────────────────────────
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn team_store_roundtrip() {
+        let tmp = tempfile::TempDir::new().unwrap();
+        let dir = tmp.path().to_path_buf();
+
+        let mut store = TeamStore { teams: HashMap::new() };
+        store.teams.insert("t1".into(), Team {
+            id: "t1".into(),
+            name: "Alpha".into(),
+            created_by: "u1".into(),
+            members: vec!["u1".into(), "u2".into()],
+            created_at: "2026-01-01".into(),
+        });
+        store.save(&dir).unwrap();
+
+        let loaded = TeamStore::load(&dir);
+        assert_eq!(loaded.teams.len(), 1);
+        assert_eq!(loaded.teams["t1"].name, "Alpha");
+        assert_eq!(loaded.teams["t1"].members.len(), 2);
+    }
+
+    #[test]
+    fn teams_for_user_filters_correctly() {
+        let mut store = TeamStore { teams: HashMap::new() };
+        store.teams.insert("t1".into(), Team {
+            id: "t1".into(), name: "A".into(), created_by: "u1".into(),
+            members: vec!["u1".into(), "u2".into()], created_at: "".into(),
+        });
+        store.teams.insert("t2".into(), Team {
+            id: "t2".into(), name: "B".into(), created_by: "u3".into(),
+            members: vec!["u3".into()], created_at: "".into(),
+        });
+
+        assert_eq!(store.teams_for_user("u1").len(), 1);
+        assert_eq!(store.teams_for_user("u2").len(), 1);
+        assert_eq!(store.teams_for_user("u3").len(), 1);
+        assert_eq!(store.teams_for_user("u99").len(), 0);
+    }
+}
--- a/cthulu-backend/api/mod.rs
+++ b/cthulu-backend/api/mod.rs
@@ -10,6 +10,7 @@ pub mod middleware;
 pub mod prompts;
 mod routes;
 pub mod scheduler;
+pub mod teams;
 pub mod templates;
 pub mod user_context;
 
@@ -340,6 +341,8 @@ pub struct AppState {
     pub jwt_secret: Arc<String>,
     /// In-memory user store (email/password accounts).
     pub user_store: Arc<RwLock<local_auth::UserStore>>,
+    /// In-memory team store.
+    pub team_store: Arc<RwLock<teams::TeamStore>>,
 }
 
 impl AppState {
--- a/cthulu-backend/api/routes.rs
+++ b/cthulu-backend/api/routes.rs
@@ -58,6 +58,7 @@ fn api_router() -> Router<AppState> {
         .merge(super::hooks::router())
         .merge(super::dashboard::router())
         .merge(super::local_auth::router())
+        .merge(super::teams::router())
 }
 
 async fn not_found(req: axum::extract::Request) -> impl IntoResponse {
--- a/cthulu-backend/main.rs
+++ b/cthulu-backend/main.rs
@@ -374,6 +374,9 @@ async fn run_server(start_disabled: bool) -> Result<(), Box<dyn Error>> {
         user_store: Arc::new(tokio::sync::RwLock::new(
             crate::api::local_auth::UserStore::load(&base_dir),
         )),
+        team_store: Arc::new(tokio::sync::RwLock::new(
+            crate::api::teams::TeamStore::load(&base_dir),
+        )),
     };
 
     // Start file change watcher (keeps caches in sync with external edits)


saltyskip:

_Reviewing PR #120 — this may take a few minutes._

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,438 @@
...
+    };
+    drop(store);
+
+    let valid = bcrypt::verify(&body.password, &user.password_hash).unwrap_or(false);
+    if !valid {
+        return (
+            StatusCode::UNAUTHORIZED,
+            Json(json!({ "error": "Invalid email or password" })),
+        );
+    }
```
saltyskip, line 370: **Bug: blocking bcrypt on async runtime.** `bcrypt::verify` is CPU-intensive (same as `bcrypt::hash`), but unlike `signup` where you correctly use `spawn_blocking` for hashing, here `verify` runs directly on the Tokio async runtime. This will block the executor thread for the duration of the bcrypt computation (~100-300ms at cost 12).

```rust
let hash = user.password_hash.clone();
let valid = match tokio::task::spawn_blocking(move || bcrypt::verify(&password, &hash)).await {
    Ok(Ok(v)) => v,
    _ => false,
};
```

cthulu-backend/api/clerk_auth.rs
```diff
@@ -0,0 +1,438 @@
...
+            Ok(contents) => serde_json::from_str(&contents).unwrap_or_default(),
+            Err(_) => HashMap::new(),
+        };
+        Self { users }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("users.json");
+        let json = serde_json::to_string_pretty(&self.users)?;
+        std::fs::write(path, json)
```
saltyskip: **Non-atomic write.** CLAUDE.md rule #9 requires atomic persistence via temp file + rename. Both `UserStore::save` and `TeamStore::save` use direct `std::fs::write`, which can produce a corrupted/truncated `users.json` if the process crashes mid-write. Should use the same temp + rename pattern as `save_sessions`:

```rust
std::fs::write(&tmp, json)?;
std::fs::rename(tmp, path)?;
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+            Ok(contents) => serde_json::from_str(&contents).unwrap_or_default(),
+            Err(_) => HashMap::new(),
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
```
saltyskip, line 45: Same non-atomic write issue as `UserStore::save` — should use temp file + rename.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+
+async fn remove_member(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path((id, user_id)): Path<(String, String)>,
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
```
saltyskip, line 184: **Edge case: self-removal deletes team without confirmation.** A member can remove themselves (`user_id == auth.user_id`), and if they're the last member, the team is silently auto-deleted (line 185-186). This seems like it could be surprising — consider either (a) preventing self-removal (use a "leave team" endpoint instead), or (b) at minimum returning a response that indicates the team was deleted so the client can react appropriately.

cthulu-backend/api/user_context.rs
```diff
@@ -0,0 +1,70 @@
...
+    #[test]
+    fn user_data_dir_path_construction() {
+        let base = PathBuf::from("/tmp/cthulu");
+        let result = base.join("users").join("user_abc");
+        assert_eq!(result, PathBuf::from("/tmp/cthulu/users/user_abc"));
+    }
+
+    #[test]
+    fn user_key_format() {
+        let result = format!("{}::{}", "user_abc", "agent::123");
```
saltyskip: **Tests don't exercise the actual functions.** `user_key_formats_correctly` manually reconstructs the format string instead of calling `user_key()`. `ensure_dirs_creates_all_subdirs` manually calls `create_dir_all` instead of calling `ensure_user_dirs()`. These tests verify nothing about the real code — they'll pass even if the functions are completely wrong.

The comment on line 50 explains why (AppState is hard to construct), but it would be better to either build a minimal test AppState or just remove these tests until that's possible. As-is they provide false confidence.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,438 @@
...
+/// Load or generate the JWT signing secret.
+pub fn load_or_create_jwt_secret(data_dir: &PathBuf) -> String {
+    let path = data_dir.join("jwt_secret");
+    match std::fs::read_to_string(&path) {
+        Ok(secret) if !secret.trim().is_empty() => secret.trim().to_string(),
+        _ => {
+            let secret = uuid::Uuid::new_v4().to_string();
+            let _ = std::fs::write(&path, &secret);
+            secret
+        }
```
saltyskip, line 118: **Silent failure if secret can't be persisted.** `let _ = std::fs::write(...)` discards the error. If the data dir doesn't exist or isn't writable, the secret is generated but never saved — next restart will generate a different secret, invalidating all existing JWTs. Consider at least logging a warning.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+
+async fn get_team(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path(id): Path<String>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.team_store.read().await;
+    match store.teams.get(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            let user_store = state.user_store.read().await;
```
saltyskip, line 115: **Holding two RwLock read guards simultaneously.** `team_store.read()` is held (line 112) while also acquiring `user_store.read()` here. With `tokio::sync::RwLock` this won't deadlock for read-read, but if any future code acquires these locks in the opposite order with a write lock, you'll get a deadlock. Consider dropping the team_store read guard (clone the team data you need) before acquiring user_store, to establish a consistent lock ordering.

saltyskip (changes requested):
## 🤖 Cthulu Review

The teams module is clean and well-structured — good separation of concerns, proper auth gating, and solid test coverage for the store layer.

**Must fix before merge:**

1. **`bcrypt::verify` blocks the async runtime** (`clerk_auth.rs:319`) — `signup` correctly uses `spawn_blocking` for hashing, but `login` calls `bcrypt::verify` directly on the Tokio executor. At bcrypt cost 12 this blocks for ~100-300ms per login. Wrap in `spawn_blocking` to match `signup`.

2. **Non-atomic JSON writes** (`UserStore::save`, `TeamStore::save`) — both use bare `std::fs::write`, which can corrupt the file on crash. The project's own CLAUDE.md rule #9 requires temp file + rename for persistence. Apply the same pattern used in `save_sessions`.

**Should fix:**

3. **`user_context.rs` tests don't call the actual functions** — they manually replicate the logic, so they'll pass even if the real code is broken. Either build a test AppState or remove them.

4. **JWT secret write failure is silently ignored** (`clerk_auth.rs:85`) — if the write fails, a new secret is generated on every restart, invalidating all tokens. At minimum log a warning.

5. **`remove_member` self-removal silently deletes team** — consider returning a signal in the response when auto-deletion happens, or preventing self-removal in favor of an explicit "leave" action.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #120 (previous HEAD: `0f37493`, new HEAD: `1f3c642`)_

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
```
saltyskip, line 47: **Bug: Non-atomic save will corrupt data on crash.**

`UserStore::save()` in `clerk_auth.rs` correctly uses temp-file + rename (matching CLAUDE.md rule #9), but `TeamStore::save()` writes directly with `std::fs::write`. A crash or power loss mid-write will corrupt `teams.json`.

Fix:
```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        members: vec![auth.user_id.clone()],
+        created_at: chrono::Utc::now().to_rfc3339(),
+    };
+
+    let mut store = state.team_store.write().await;
+    let response = json!({ "team": &team });
+    store.teams.insert(team.id.clone(), team);
+    let _ = store.save(&state.data_dir);
+
+    (StatusCode::CREATED, Json(response))
```
saltyskip, line 104: Save errors are silently discarded with `let _ = store.save(...)`. If the write fails, in-memory state diverges from disk — all changes are lost on restart with no indication to the caller.

At minimum, log the error:
```rust
    tracing::error!("Failed to save teams: {e}");
}
```

Same issue on lines 122, 139, and 157.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+
+async fn add_member(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path(id): Path<String>,
+    Json(body): Json<AddMemberRequest>,
+) -> (StatusCode, Json<Value>) {
+    let email = body.email.trim().to_lowercase();
+
+    let user_store = state.user_store.read().await;
```
saltyskip, line 153: Minor: `remove_member` lets any member remove any other member (including the team creator). The PR description says "no roles — all members equal" so this is by design, but consider whether the creator should be protected from removal, or at least documented as intentional in a code comment.

cthulu-backend/api/user_context.rs
```diff
@@ -0,0 +1,76 @@
...
+    #[test]
+    fn user_key_formats_correctly() {
+        // Verify the format string matches what user_key produces
+        let auth = test_auth("user_abc");
+        let expected = format!("{}::{}", auth.user_id, "agent::123");
+        assert_eq!(expected, "user_abc::agent::123");
+    }
+
+    #[test]
+    fn ensure_dirs_creates_all_subdirs() {
```
saltyskip: These tests don't actually call `user_key()` or `ensure_user_dirs()` — they duplicate the logic manually. The `user_key` test just tests `format!()`, and the `ensure_dirs` test just calls `create_dir_all` directly.

Consider either:
1. Building a minimal test `AppState` to test the real functions, or
2. Removing these tests since they don't exercise the actual code paths.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Must fix before merge:**

1. **`TeamStore::save()` is not atomic** — `UserStore::save()` in the same PR correctly uses temp-file + rename (per CLAUDE.md rule #9), but `TeamStore::save()` uses bare `std::fs::write()`. A crash mid-write corrupts `teams.json`. Copy the same pattern from `UserStore::save()`.

2. **Save errors silently discarded** — All four `let _ = store.save(...)` calls mean disk failures go unnoticed. In-memory state diverges and changes are lost on restart. At minimum log the errors.

**Suggestions (non-blocking):**

- `user_context.rs` tests don't call the actual functions — they duplicate logic manually. Either build a test `AppState` or remove the tests.
- `remove_member` allows any member to remove the creator (and auto-deletes empty teams). This is fine per "no roles" design, but a code comment would clarify intent.

Overall the structure is clean — CRUD handlers, membership checks, and the auth integration all look correct. Just need the atomic save fix.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #120 (previous HEAD: `1f3c642`, new HEAD: `3caa22e`)_

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+            Err(_) => HashMap::new(),
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
```
saltyskip, line 46: **Bug: Non-atomic save violates project rule #9.** `TeamStore::save` uses bare `std::fs::write`, which can corrupt the file on a crash mid-write. The sibling `UserStore::save` in `clerk_auth.rs` correctly does temp file + rename. This should match:

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+            }
+            team.members.push(target_id);
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
```
saltyskip, line 175: **Design: Any member can remove any other member (including the creator), and the team auto-deletes when empty.** This means a single member can grief-delete an entire team by removing everyone else then themselves. Consider either:
- Only allowing `created_by` to remove others, or
- Preventing removal of the last member / the creator

At minimum, document this as intentional if it is.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    State(state): State<AppState>,
+    Path(id): Path<String>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.team_store.read().await;
+    match store.teams.get(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            let user_store = state.user_store.read().await;
+            let members: Vec<Value> = team.members.iter().map(|uid| {
+                let user = user_store.users.values().find(|u| u.id == *uid);
+                json!({
```
saltyskip, line 118: **Potential lock ordering concern:** This handler holds `team_store.read()` (line 113) and then acquires `user_store.read()` (line 117). If any future code acquires these locks in the opposite order, you'll deadlock. Consider dropping the team store read lock before acquiring the user store lock (clone what you need from the team first), or document the required lock ordering.

cthulu-backend/api/user_context.rs
```diff
@@ -0,0 +1,78 @@
...
+    // We test the path/key construction logic directly; full integration tests
+    // would need a test AppState builder (future improvement).
+
+    #[test]
+    fn user_key_formats_correctly() {
+        // Verify the format string matches what user_key produces
+        let auth = test_auth("user_abc");
+        let expected = format!("{}::{}", auth.user_id, "agent::123");
+        assert_eq!(expected, "user_abc::agent::123");
+    }
```
saltyskip, line 82: **Tests don't test actual code.** This test manually does `format!("{}::{}", ...)` instead of calling `user_key()`. The `ensure_dirs_creates_all_subdirs` test below similarly manually creates dirs instead of calling `ensure_user_dirs()`. These tests verify string formatting works, not that your functions work. If building a test `AppState` is too heavy right now, consider at least adding a `// TODO: test actual functions once test AppState builder exists` and removing these misleading tests.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,451 @@
...
+        Ok(secret) if !secret.trim().is_empty() => secret.trim().to_string(),
+        _ => {
+            // 256 bits of randomness: two UUID v4s concatenated (each has 122 bits)
+            let secret = format!("{}{}", uuid::Uuid::new_v4(), uuid::Uuid::new_v4())
+                .replace('-', "");
+            let _ = std::fs::write(&path, &secret);
+            secret
+        }
+    }
+}
```
saltyskip, line 120: **Minor: JWT secret entropy comment is slightly inaccurate.** UUIDv4 provides ~122 random bits each (6 bits are version/variant), so two give ~244 bits, not 256. Practically fine, but the comment could mislead a future security reviewer. Consider using `rand` to generate exactly 32 random bytes encoded as hex, or just fix the comment.

saltyskip (changes requested):
## 🤖 Cthulu Review

The teams module is clean and well-structured — flat membership, simple CRUD, good test coverage of the store layer. Two issues should be fixed before merge:

**Must fix:**
- **`TeamStore::save` is not atomic** — it uses bare `std::fs::write` which can corrupt `teams.json` on a crash. The `UserStore::save` right next to it correctly does temp file + rename (project rule #9). This is a one-line fix.

**Should address:**
- **Any member can remove any other member** — including the team creator. Combined with auto-delete-on-empty, one rogue member can wipe a team. If this is intentional "no roles" design, document it; otherwise, add a guard.

**Minor / non-blocking:**
- `user_context.rs` tests don't call the actual functions — they manually replicate the logic. Consider removing or marking as TODOs.
- `get_team` holds two read locks simultaneously — fine today, but document the lock ordering or restructure to avoid future deadlock risk.
- JWT secret entropy comment says "256 bits" but UUIDv4 pairs give ~244 bits.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #120 (previous HEAD: `3caa22e`, new HEAD: `8771ab0`)_

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+            Ok(contents) => serde_json::from_str(&contents).unwrap_or_default(),
+            Err(_) => HashMap::new(),
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
```
saltyskip, line 45: **Bug: Non-atomic save — crash can corrupt `teams.json`**

This uses `std::fs::write(path, json)` directly, which is not crash-safe. If the process is killed mid-write, `teams.json` will be left partially written (corrupted). On next startup, `TeamStore::load` will hit `unwrap_or_default()` and silently lose all teams.

The `UserStore::save` in the same PR correctly uses temp+rename:
```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

This is also required by project rule #9 (atomic YAML/JSON persistence via temp+rename).

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        name,
+        created_by: auth.user_id.clone(),
+        members: vec![auth.user_id.clone()],
+        created_at: chrono::Utc::now().to_rfc3339(),
+    };
+
+    let mut store = state.team_store.write().await;
+    let response = json!({ "team": &team });
+    store.teams.insert(team.id.clone(), team);
+    let _ = store.save(&state.data_dir);
```
saltyskip, line 102: **Silent error suppression** — `let _` discards the save error entirely, with no log. If the disk is full or permissions are wrong, users will create teams that appear to work but are lost on restart.

The auth module handles this properly:
```rust
    tracing::error!(error = %e, "failed to persist team store");
}
```

Same issue on lines 168 and 188.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+
+async fn remove_member(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path((id, user_id)): Path<(String, String)>,
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
```
saltyskip, line 184: **`remove_member` silently succeeds for non-existent members** — If `user_id` is not in the members list, `retain` is a no-op but the handler still returns `{ "ok": true }` and triggers a save. Consider returning 404 when the target user isn't actually a member:

```rust
if !team.members.contains(&user_id) {
    return (StatusCode::NOT_FOUND, Json(json!({ "error": "User is not a member" })));
}
team.members.retain(|m| m != &user_id);
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+async fn remove_member(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path((id, user_id)): Path<(String, String)>,
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
```
saltyskip, line 183: **Self-removal deletes the team with no warning** — A member can remove themselves (`user_id == auth.user_id`), and if they're the last member, the team is silently deleted (line 185-187). Consider either:
1. Preventing self-removal if you're the last member, or
2. Returning a response that indicates the team was deleted (e.g. `{ "ok": true, "team_deleted": true }`) so the frontend can react.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Must fix before merge:**

1. **`TeamStore::save` is not atomic** — Direct `std::fs::write` can corrupt `teams.json` on crash. Use temp+rename like `UserStore::save` does. This violates project rule #9.

2. **Save errors silently discarded** — All three team mutation handlers use `let _ = store.save(...)`, losing the error. At minimum log with `tracing::error!` (as `local_auth.rs` does for the user store).

3. **`remove_member` returns 200 for non-existent members** — `retain` is a no-op if the user isn't in the list, but the handler still returns success and writes to disk.

**Worth considering:**

4. Self-removal as last member silently deletes the team — the caller gets `{ "ok": true }` with no indication the team no longer exists.

The auth module (`local_auth.rs`, `user_context.rs`, config changes) looks solid — good use of `spawn_blocking` for bcrypt, proper TOCTOU prevention on signup, and the JWT secret generation with file permissions is well done.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #120 (previous HEAD: `8771ab0`, new HEAD: `0fd436f`)_

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+            Err(_) => HashMap::new(),
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
```
saltyskip, line 46: **Bug: Non-atomic save — risks data corruption on crash.**

`UserStore::save` in `local_auth.rs` correctly uses temp-file + rename for crash-safe writes, but `TeamStore::save` uses `std::fs::write` directly. If the process crashes mid-write, `teams.json` will be truncated/corrupt and all team data is lost.

This also violates the project convention (CLAUDE.md rule #9: atomic persistence via temp + rename).

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    let mut store = state.team_store.write().await;
+    let response = json!({ "team": &team });
+    store.teams.insert(team.id.clone(), team);
+    let _ = store.save(&state.data_dir);
+
+    (StatusCode::CREATED, Json(response))
+}
+
+async fn get_team(
```
saltyskip, line 107: **Silent data loss: save errors are swallowed.**

`let _ = store.save(...)` discards the error. If the write fails, the in-memory state diverges from disk and the user gets a 201 Created for a team that won't survive a restart. Same pattern repeats in `add_member` (line 160) and `remove_member` (line 176).

`local_auth.rs` at least logs errors with `tracing::error!`. Consider doing the same here, or returning a 500.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            if team.members.contains(&target_id) {
+                return (StatusCode::CONFLICT, Json(json!({ "error": "Already a member" })));
+            }
+            team.members.push(target_id);
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
```
saltyskip, line 170: **Logic issue: any member can remove any other member, including the creator.**

Since there are no roles (by design), any member can remove any other member. Combined with auto-delete-on-empty (line 172-174), a single malicious member could remove everyone else and then themselves, deleting the team entirely.

This may be intentional for the "flat, no roles" model, but worth a conscious decision. At minimum, consider preventing removal of `created_by` unless they're the one leaving.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        None => return (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
+    };
+    drop(user_store);
+
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            if team.members.contains(&target_id) {
+                return (StatusCode::CONFLICT, Json(json!({ "error": "Already a member" })));
+            }
```
saltyskip, line 166: **Bug: a non-member can be "removed" without error.**

If `user_id` in the path doesn't match any member, `retain` is a no-op but the handler returns 200 OK. Should check if the user was actually a member before returning success — otherwise the client can't distinguish between a successful removal and a no-op.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Must fix before merge:**

1. **Non-atomic `TeamStore::save`** — uses `std::fs::write` directly instead of temp-file + rename. A crash mid-write corrupts `teams.json` and loses all team data. `UserStore::save` in the same PR does this correctly — just copy the pattern.

2. **Silent save errors** — `let _ = store.save(...)` in `create_team`, `add_member`, and `remove_member` swallows I/O errors. The user gets a success response for a team that won't survive a restart. At minimum log with `tracing::error!` (as `local_auth.rs` already does).

3. **Removing a non-member returns 200 OK** — `remove_member` doesn't check whether the target `user_id` was actually in `members` before returning success. The `retain` call is a silent no-op.

**Worth discussing:**

- Any member can remove any other member (including the creator), and auto-delete-on-empty means a single member can destroy the entire team. This may be fine for the "flat, no roles" model, but should be a conscious decision.

Overall the structure is clean and well-tested. The three bugs above are straightforward fixes.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #120 (previous HEAD: `0fd436f`, new HEAD: `6d31fb5`)_

cthulu-backend/api/teams.rs
saltyskip, line 45: **Bug: Non-atomic save — violates project rule #9.**

`std::fs::write` is not crash-safe. If the process dies mid-write, `teams.json` will be truncated/corrupt. `UserStore::save` in `local_auth.rs` correctly uses temp+rename. This should do the same:

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+/// In-memory team store backed by `{data_dir}/teams.json`.
+pub struct TeamStore {
+    pub teams: HashMap<String, Team>,
+}
+
+impl TeamStore {
+    pub fn load(data_dir: &PathBuf) -> Self {
+        let path = data_dir.join("teams.json");
+        let teams = match std::fs::read_to_string(&path) {
+            Ok(contents) => serde_json::from_str(&contents).unwrap_or_default(),
```
saltyskip, line 36: **Bug: Parse errors silently swallowed.**

`unwrap_or_default()` will discard a corrupt `teams.json` without any warning. Compare with `UserStore::load` in `local_auth.rs` which properly logs `"corrupt users.json"`. A user who has a malformed file will lose all teams on the next save with no indication of what happened.

Suggest matching the `UserStore::load` pattern:
```rust
Ok(contents) => match serde_json::from_str(&contents) {
    Ok(parsed) => parsed,
    Err(e) => {
        tracing::error!(path = %path.display(), error = %e,
            "corrupt teams.json — starting with empty store");
        HashMap::new()
    }
},
```

cthulu-backend/api/teams.rs
saltyskip, line 102: **Bug: Save failure silently ignored.** `let _ = store.save(...)` discards IO errors. The team is already in memory, so on the next mutation it could be persisted — but if the server restarts before that, the team is lost.

At minimum log the error, consistent with `UserStore` in signup handler:
```rust
    tracing::error!(error = %e, "failed to persist team store");
}
```

Same issue on lines 168 and 188.

cthulu-backend/api/teams.rs
saltyskip, line 184: **Design concern: Users can remove themselves from a team.**

When `user_id == auth.user_id`, the caller removes themselves. If they're the last member, the team is auto-deleted (line 185-186). This is documented behavior, but there's no confirmation or distinct response — the caller gets `{"ok": true}` with no indication the team was deleted.

Consider returning something like `{"ok": true, "team_deleted": true}` when the team is removed so the frontend can react appropriately.

saltyskip (changes requested):
## 🤖 Cthulu Review

The teams module is well-structured — clean separation, proper auth gating, good test coverage. However, there are a few issues that should be fixed before merging:

### Must fix

1. **`TeamStore::save` is not atomic** — Uses bare `std::fs::write` instead of temp+rename. This violates the project's rule #9 (atomic persistence) and risks data corruption on crash. `UserStore::save` in the same PR does this correctly — just copy the pattern.

2. **`TeamStore::load` silently swallows corrupt JSON** — `unwrap_or_default()` means a corrupt `teams.json` will be replaced with an empty store on next save, with no log. `UserStore::load` handles this properly with an error log — match that pattern.

3. **Save errors silently discarded** — Three instances of `let _ = store.save(...)` in handlers. If disk write fails, the in-memory state silently diverges. At minimum log the error (like the signup handler does for `UserStore`).

### Suggestion

4. **`remove_member` returns `{"ok": true}` even when it auto-deletes the team** — When the last member is removed, the team is deleted. The response gives no indication, so the frontend can't distinguish "member removed" from "team gone". Consider returning `{"ok": true, "team_deleted": true}` or similar.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #120 (previous HEAD: `6d31fb5`, new HEAD: `bd06e12`)_

cthulu-backend/api/teams.rs
saltyskip, line 45: **Bug: Non-atomic save — data loss risk on crash.**

`TeamStore::save` does a direct `std::fs::write`, but `UserStore::save` in the same PR correctly uses temp file + rename for crash safety. This also violates CLAUDE.md rule #9 ("Atomic YAML persistence: Always write sessions via temp file + rename").

If the process crashes mid-write, `teams.json` will be truncated/corrupt and `TeamStore::load` will silently discard all teams (see next comment).

Fix — match what `UserStore::save` does:
```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
saltyskip, line 36: **Bug: Silent data loss on corrupt `teams.json`.**

`unwrap_or_default()` silently replaces corrupt data with an empty HashMap. The next `save()` call will then overwrite the corrupt file with `{}`, permanently losing all teams.

`UserStore::load` in the same PR correctly logs an error for this case. Should do the same here:
```rust
Ok(contents) => match serde_json::from_str(&contents) {
    Ok(parsed) => parsed,
    Err(e) => {
        tracing::error!(path = %path.display(), error = %e,
            "corrupt teams.json — starting with empty store");
        HashMap::new()
    }
},
```

cthulu-backend/api/teams.rs
saltyskip, line 102: **Issue: Silent save failures in all three team mutation handlers.**

`let _ = store.save(...)` discards write errors silently. If the disk is full or permissions wrong, the in-memory state will diverge from disk — changes are lost on restart with no indication.

`local_auth.rs` line 329 correctly logs save failures. All three instances (lines 102, 168, 188) should at minimum log:
```rust
    tracing::error!(error = %e, "failed to persist team store");
}
```

cthulu-backend/api/teams.rs
saltyskip, line 184: **Design: Any member can remove any other member (including the creator).**

The only auth check is "is the requester a member". This means user A can kick user B (the creator) out of a team. The PR description says "no roles — all members equal" so this may be intentional, but it's worth calling out explicitly since it can lead to surprising behavior (e.g., the last person added can remove everyone else and become the sole owner of a team they didn't create).

Also: a member can remove *themselves* and if they're the last member, the team is auto-deleted. Consider whether self-removal should be a separate "leave" action that warns when you're the last member.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,487 @@
...
+) -> (StatusCode, Json<Value>) {
+    let email = body.email.trim().to_lowercase();
+
+    let store = state.user_store.read().await;
+    let user = store.find_by_email(&email).cloned();
+    drop(store);
+
+    // Constant-time: always run bcrypt verify to prevent user enumeration via timing.
+    // If user doesn't exist, verify against a dummy hash (same cost as real verify).
+    let dummy_hash = "$2b$12$000000000000000000000u2a5OJr0FkDxcMkGCuaLxMPqOIZJcMK";
```
saltyskip, line 357: **Minor: Dummy bcrypt hash may not fully prevent timing-based user enumeration.**

Good thinking using a constant-time approach. However, if `$2b$12$000000000000000000000u...` is not a structurally valid bcrypt hash (all-zero salt bytes may be rejected by some implementations), `bcrypt::verify` might return `Err` immediately rather than doing the full cost-12 computation. This would make the non-existent-user path measurably faster.

Safer: generate a real dummy hash once at startup:
```rust
// In AppState init or as a lazy_static:
let dummy_hash = bcrypt::hash("dummy_password_placeholder", 12).unwrap();
```
Then reference it in the login handler.

saltyskip (changes requested):
## 🤖 Cthulu Review

Good structure overall — clean separation of concerns, proper TOCTOU-safe signup, constant-time login, and solid test coverage on both stores.

**3 issues to fix before merge:**

1. **`TeamStore::save` is not atomic** — direct `std::fs::write` instead of temp+rename. Crash mid-write will corrupt `teams.json`. The `UserStore::save` in the same PR does this correctly — just copy the pattern.

2. **`TeamStore::load` silently swallows parse errors** — `unwrap_or_default()` on corrupt JSON means the next save overwrites the file with `{}`, permanently losing all teams. Log the error like `UserStore::load` does.

3. **Silent save failures in all team handlers** — `let _ = store.save(...)` discards errors. At minimum log them, as `local_auth.rs` already does for user store.

**1 minor concern (non-blocking):**
- The dummy bcrypt hash in login may not actually produce constant-time behavior if the hash is structurally invalid. Consider generating a real hash at startup.

All three blocking issues are one-line fixes that bring `TeamStore` in line with `UserStore` patterns already established in this same PR.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #120 (previous HEAD: `bd06e12`, new HEAD: `dc94967`)_

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+            Err(_) => HashMap::new(),
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
```
saltyskip, line 42: **Bug: Non-atomic save — violates project rule #9 and risks data loss.**

`TeamStore::save()` uses direct `std::fs::write()`, but `UserStore::save()` (in the same PR) correctly uses temp-file + rename for crash safety. If the process crashes mid-write, `teams.json` will be truncated/corrupt.

This also makes `load()` silently discard the corrupt file via `unwrap_or_default()` on line 37, so you'd lose all teams with no log.

Fix both:
```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

And in `load()`, log parse errors like `UserStore` does instead of silently discarding:
```rust
Ok(contents) => match serde_json::from_str(&contents) {
    Ok(parsed) => parsed,
    Err(e) => {
        tracing::error!("corrupt teams.json: {e}");
        HashMap::new()
    }
},
```

cthulu-backend/api/teams.rs
saltyskip, line 107: Save errors silently swallowed with `let _ = store.save(...)`. This pattern repeats in `add_member` (line 163) and `remove_member` (line 179). Compare with `UserStore` in `local_auth.rs` which logs:
```rust
    tracing::error!(error = %e, "failed to persist team store");
}
```
Silent failures here mean a user creates a team, gets a 201 back, but the team disappears on restart.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+            }
+            team.members.push(target_id);
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
```
saltyskip, line 175: **`remove_member` doesn't verify the target is actually a member.** `retain()` is a no-op for non-members, so this returns `200 OK` with `{"ok": true}` even when removing a user_id that was never in the team. Should return 404 or at least check membership first:
```rust
if !team.members.contains(&user_id) {
    return (StatusCode::NOT_FOUND, Json(json!({ "error": "User is not a member" })));
}
team.members.retain(|m| m != &user_id);
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+async fn remove_member(
```
saltyskip, line 176: **Self-removal auto-deletes the team with no safeguard.** If a user removes themselves and they're the last member, the team is silently deleted. This could cause data loss if done accidentally. Consider either:
- Preventing self-removal ("use DELETE /teams/{id} instead")
- Or at minimum, returning a response indicating the team was deleted so the client can react

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+pub struct TeamStore {
+    pub teams: HashMap<String, Team>,
+}
+
+impl TeamStore {
+    pub fn load(data_dir: &PathBuf) -> Self {
+        let path = data_dir.join("teams.json");
+        let teams = match std::fs::read_to_string(&path) {
+            Ok(contents) => serde_json::from_str(&contents).unwrap_or_default(),
+            Err(_) => HashMap::new(),
```
saltyskip, line 37: Parse errors silently discarded via `unwrap_or_default()`. Unlike `UserStore::load()` which distinguishes parse errors from missing files and logs corruption, this treats a corrupt `teams.json` the same as a missing one — all teams vanish silently.

saltyskip (changes requested):
## 🤖 Cthulu Review

**TeamStore persistence is not crash-safe and swallows errors.** The `UserStore` in the same PR correctly uses temp+rename (matching project rule #9), but `TeamStore` uses direct `std::fs::write` and `let _ =` to discard save errors. This means:
- A crash during save corrupts `teams.json`
- `load()` then silently discards the corrupt file (`unwrap_or_default`)
- Save failures in handlers return 200/201 to the client, but data is lost on restart

Other issues:
- `remove_member` returns success even when the target isn't a member (no-op `retain`)
- Self-removal silently auto-deletes the team with no indication to the client

The auth module (`local_auth.rs`, `user_context.rs`) looks solid — good patterns with bcrypt on `spawn_blocking`, timing-safe login, TOCTOU prevention on signup, and JWT secret file permissions. The teams module just needs to follow the same patterns already established in `UserStore`.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #120 (previous HEAD: `dc94967`, new HEAD: `81564e1`)_

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
```
saltyskip, line 44: **Non-atomic write violates project convention.** `local_auth::UserStore::save` already uses the temp-file + rename pattern (and CLAUDE.md rule #9 requires it). `TeamStore::save` should do the same to avoid data corruption on crash:

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(tmp, path)
}
```

cthulu-backend/api/teams.rs
saltyskip, line 102: **Silent save failure.** `let _ = store.save(...)` discards the error. If persistence fails, in-memory state silently diverges from disk. At minimum log the error:

```rust
    tracing::error!("Failed to save teams: {e}");
}
```

Same issue on lines 150, 166, and 175.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    let user_store = state.user_store.read().await;
+    let target_user = user_store.find_by_email(&email);
+    let target_id = match target_user {
+        Some(u) => u.id.clone(),
+        None => return (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
+    };
+    drop(user_store);
+
+    let mut store = state.team_store.write().await;
```
saltyskip, line 161: **Returns 200 OK even when target user isn't a member.** `retain` is a no-op if `user_id` doesn't match any member, but the handler still returns success and triggers a save. Consider checking membership first:

```rust
if !team.members.contains(&user_id) {
    return (StatusCode::NOT_FOUND, Json(json!({ "error": "User is not a member" })));
}
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    (StatusCode::CREATED, Json(response))
+}
+
+async fn get_team(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path(id): Path<String>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.team_store.read().await;
+    match store.teams.get(&id) {
```
saltyskip, line 113: **O(n) linear scan per member.** `user_store.users` is keyed by email, so `users.values().find(|u| u.id == *uid)` is a linear scan for every member. For teams with many members this adds up. Consider adding an `id -> email` index to `UserStore`, or at least a `find_by_id` helper to keep the concern in one place.

saltyskip (changes requested):
## 🤖 Cthulu Review

Clean implementation overall — the flat team model, auth checks, and auto-delete-on-empty are solid. Four issues to address before merging:

1. **Non-atomic persistence** — `TeamStore::save` uses `std::fs::write` directly, which can corrupt `teams.json` on crash. `UserStore::save` already uses the temp-file + rename pattern per project convention (CLAUDE.md rule #9). `TeamStore` should match.

2. **Silent save failures** — `let _ = store.save(...)` in all four mutation handlers discards I/O errors. At minimum, log them so data loss doesn't go unnoticed.

3. **`remove_member` returns 200 for non-members** — If the target `user_id` isn't in the team, `retain` is a no-op but the handler returns success and writes to disk. Should return 404 instead.

4. **O(n) user lookup by ID** — `get_team` does a linear scan through all users for each member since `UserStore` is keyed by email. Fine for now but worth a `find_by_id` helper to contain the concern.

Items 1-3 are blocking; item 4 is a suggestion.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #120 (previous HEAD: `81564e1`, new HEAD: `24335ac`)_

cthulu-backend/api/teams.rs
saltyskip, line 44: **Bug: Non-atomic save risks data corruption on crash.**

`UserStore::save()` uses the temp-file + rename pattern (crash-safe), but `TeamStore::save()` uses direct `std::fs::write`. If the process crashes mid-write, `teams.json` will be truncated/corrupt and `load()` will silently discard all teams.

Suggested fix:
```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
saltyskip, line 37: **Data loss: corrupt JSON silently discarded.**

`unwrap_or_default()` means if `teams.json` exists but is corrupt (e.g. partial write from the non-atomic save above), all teams are silently lost on next server start. `UserStore::load()` logs `tracing::error!` for this case — `TeamStore` should do the same.

```rust
let teams = match std::fs::read_to_string(&path) {
    Ok(contents) => match serde_json::from_str(&contents) {
        Ok(parsed) => parsed,
        Err(e) => {
            tracing::error!(path = %path.display(), error = %e,
                "corrupt teams.json — starting with empty store");
            HashMap::new()
        }
    },
    Err(e) if e.kind() != std::io::ErrorKind::NotFound => {
        tracing::error!(path = %path.display(), error = %e, "failed to read teams.json");
        HashMap::new()
    }
    Err(_) => HashMap::new(),
};
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        members: vec![auth.user_id.clone()],
+        created_at: chrono::Utc::now().to_rfc3339(),
+    };
+
+    let mut store = state.team_store.write().await;
+    let response = json!({ "team": &team });
+    store.teams.insert(team.id.clone(), team);
+    let _ = store.save(&state.data_dir);
+
+    (StatusCode::CREATED, Json(response))
```
saltyskip, line 104: **Silent error: save failure ignored.**

`let _ = store.save(...)` discards any I/O error. If the disk is full or permissions change, the in-memory state will diverge from disk with no indication. At minimum, log the error:

```rust
    tracing::error!(error = %e, "failed to save teams");
}
```

This pattern appears in all three write handlers (`create_team`, `add_member`, `remove_member`).

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        Some(u) => u.id.clone(),
+        None => return (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
+    };
+    drop(user_store);
+
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            if team.members.contains(&target_id) {
+                return (StatusCode::CONFLICT, Json(json!({ "error": "Already a member" })));
```
saltyskip, line 165: **Logic issue: removing a non-existent member returns 200 OK.**

`retain` is a no-op if `user_id` isn't in the members list. This silently succeeds and triggers a save. Consider checking membership first:

```rust
if !team.members.contains(&user_id) {
    return (StatusCode::NOT_FOUND, Json(json!({ "error": "User is not a member" })));
}
team.members.retain(|m| m != &user_id);
```

saltyskip (changes requested):
## 🤖 Cthulu Review

The teams module is well-structured and the API design is clean. However, there are a few issues that should be fixed before merging:

**Must fix:**
- **Non-atomic save** — `TeamStore::save()` uses `std::fs::write` directly instead of the temp-file + rename pattern used by `UserStore::save()`. This risks data corruption on crash. Match the existing pattern.
- **Silent data loss on corrupt JSON** — `unwrap_or_default()` in `load()` silently discards corrupt `teams.json`. Add `tracing::error!` logging like `UserStore::load()` does.
- **Save errors silently ignored** — `let _ = store.save(...)` in all three write handlers. At minimum log failures so operators can diagnose disk issues.

**Should fix:**
- **`remove_member` succeeds for non-members** — `retain` is a no-op if the user_id doesn't exist, but returns 200 OK and triggers an unnecessary save. Add a membership check first.

---

_Automated review by Cthulu_

PR #121, opened 2026-03-20
feat: profile dropdown + teams UI with member management

MPRR-Pandu:
## Summary

Depends on: #119 (frontend auth) + #120 (teams backend) — merge both first.

- Profile endpoints: GET/PUT `/api/auth/me`, GET `/api/auth/users/search`
- ProfileMenu in TopBar: edit display name, create teams, manage members
- UserSearchInput: debounced search-as-you-type member picker (300ms debounce)
- Teams + profile API client functions
- Profile/team CSS styles

## Endpoints added

| GET | `/api/auth/me` | Get user profile |
| PUT | `/api/auth/me` | Update name/avatar |
| GET | `/api/auth/users/search?q=` | Search users by name/email |

## Files changed (4 files, ~395 lines)
- `clerk_auth.rs` — profile handlers (get/update/search)
- `client.ts` — profile + teams API functions
- `TopBar.tsx` — ProfileMenu + UserSearchInput components
- `styles.css` — profile + team card CSS

## Verification
- `cargo check` — 0 errors

## Dependency chain
**PR E (5/6)** — merge after #119 + #120

```
main → #117 → #118 → #119 + #120 → THIS PR E (profile + teams UI)
```

### DIFF
--- a/cthulu-backend/api/local_auth.rs
+++ b/cthulu-backend/api/local_auth.rs
@@ -7,7 +7,7 @@
 //! and all requests get a hardcoded "dev_user" identity.
 
 use axum::extract::FromRequestParts;
-use axum::routing::post;
+use axum::routing::{get, post, put};
 use axum::{Json, Router};
 use hyper::header::AUTHORIZATION;
 use hyper::http::request::Parts;
@@ -260,6 +260,8 @@ pub fn router() -> Router<AppState> {
     Router::new()
         .route("/auth/signup", post(signup))
         .route("/auth/login", post(login))
+        .route("/auth/me", get(get_profile).put(update_profile))
+        .route("/auth/users/search", get(search_users))
 }
 
 async fn signup(
@@ -387,6 +389,83 @@ async fn login(
             "user": { "id": user.id, "email": user.email }
         })),
     )
+}
+
+// ── Profile Handlers ─────────────────────────────────────────
+
+async fn get_profile(
+    auth: AuthUser,
+    axum::extract::State(state): axum::extract::State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.user_store.read().await;
+    let user = store.users.values().find(|u| u.id == auth.user_id);
+    match user {
+        Some(u) => (StatusCode::OK, Json(json!({
+            "id": u.id, "email": u.email, "name": u.name,
+            "avatar_url": u.avatar_url, "created_at": u.created_at,
+        }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
+    }
+}
+
+#[derive(Deserialize)]
+struct UpdateProfileRequest {
+    name: Option<String>,
+    avatar_url: Option<String>,
+}
+
+async fn update_profile(
+    auth: AuthUser,
+    axum::extract::State(state): axum::extract::State<AppState>,
+    Json(body): Json<UpdateProfileRequest>,
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.user_store.write().await;
+    let email_key = store.users.iter()
+        .find(|(_, u)| u.id == auth.user_id)
+        .map(|(email, _)| email.clone());
+
+    let Some(email) = email_key else {
+        return (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" })));
+    };
+    let Some(user) = store.users.get_mut(&email) else {
+        return (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" })));
+    };
+    if let Some(name) = body.name {
+        user.name = Some(name);
+    }
+    if let Some(avatar_url) = body.avatar_url {
+        user.avatar_url = Some(avatar_url);
+    }
+    let updated = user.clone();
+    let _ = store.save(&state.data_dir);
+
+    (StatusCode::OK, Json(json!({
+        "id": updated.id, "email": updated.email,
+        "name": updated.name, "avatar_url": updated.avatar_url,
+    })))
+}
+
+async fn search_users(
+    _auth: AuthUser,
+    axum::extract::State(state): axum::extract::State<AppState>,
+    axum::extract::Query(params): axum::extract::Query<std::collections::HashMap<String, String>>,
+) -> (StatusCode, Json<Value>) {
+    let query = params.get("q").map(|s| s.trim().to_lowercase()).unwrap_or_default();
+    if query.is_empty() {
+        return (StatusCode::OK, Json(json!({ "users": [] })));
+    }
+
+    let store = state.user_store.read().await;
+    let results: Vec<Value> = store.users.values()
+        .filter(|u| {
+            u.email.to_lowercase().contains(&query)
+                || u.name.as_ref().map(|n| n.to_lowercase().contains(&query)).unwrap_or(false)
+        })
+        .take(10)
+        .map(|u| json!({ "id": u.id, "email": u.email, "name": u.name }))
+        .collect();
+
+    (StatusCode::OK, Json(json!({ "users": results })))
 }
 
 // ── Tests ────────────────────────────────────────────────────
--- a/cthulu-backend/api/teams.rs
+++ b/cthulu-backend/api/teams.rs
@@ -1,0 +1,240 @@
+//! Teams: flat groups of users. No roles — all members are equal.
+//! Stored in `{data_dir}/teams.json`.
+
+use axum::extract::{Path, State};
+use axum::routing::{delete, get, post};
+use axum::{Json, Router};
+use hyper::StatusCode;
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Value};
+use std::collections::HashMap;
+use std::path::PathBuf;
+
+use crate::api::local_auth::AuthUser;
+use crate::api::AppState;
+
+// ── Team Model ───────────────────────────────────────────────
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Team {
+    pub id: String,
+    pub name: String,
+    pub created_by: String,
+    pub members: Vec<String>,
+    pub created_at: String,
+}
+
+/// In-memory team store backed by `{data_dir}/teams.json`.
+pub struct TeamStore {
+    pub teams: HashMap<String, Team>,
+}
+
+impl TeamStore {
+    pub fn load(data_dir: &PathBuf) -> Self {
+        let path = data_dir.join("teams.json");
+        let teams = match std::fs::read_to_string(&path) {
+            Ok(contents) => serde_json::from_str(&contents).unwrap_or_default(),
+            Err(_) => HashMap::new(),
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
+        self.teams.values()
+            .filter(|t| t.members.contains(&user_id.to_string()))
+            .collect()
+    }
+}
+
+// ── Routes ───────────────────────────────────────────────────
+
+pub fn router() -> Router<AppState> {
+    Router::new()
+        .route("/teams", get(list_teams).post(create_team))
+        .route("/teams/{id}", get(get_team))
+        .route("/teams/{id}/members", post(add_member))
+        .route("/teams/{id}/members/{user_id}", delete(remove_member))
+}
+
+// ── Handlers ─────────────────────────────────────────────────
+
+async fn list_teams(
+    auth: AuthUser,
+    State(state): State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.team_store.read().await;
+    let teams: Vec<&Team> = store.teams_for_user(&auth.user_id);
+    (StatusCode::OK, Json(json!({ "teams": teams })))
+}
+
+#[derive(Deserialize)]
+struct CreateTeamRequest {
+    name: String,
+}
+
+async fn create_team(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Json(body): Json<CreateTeamRequest>,
+) -> (StatusCode, Json<Value>) {
+    let name = body.name.trim().to_string();
+    if name.is_empty() {
+        return (StatusCode::BAD_REQUEST, Json(json!({ "error": "Team name required" })));
+    }
+
+    let team = Team {
+        id: uuid::Uuid::new_v4().to_string().replace('-', "_"),
+        name,
+        created_by: auth.user_id.clone(),
+        members: vec![auth.user_id.clone()],
+        created_at: chrono::Utc::now().to_rfc3339(),
+    };
+
+    let mut store = state.team_store.write().await;
+    let response = json!({ "team": &team });
+    store.teams.insert(team.id.clone(), team);
+    let _ = store.save(&state.data_dir);
+
+    (StatusCode::CREATED, Json(response))
+}
+
+async fn get_team(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path(id): Path<String>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.team_store.read().await;
+    match store.teams.get(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            let user_store = state.user_store.read().await;
+            let members: Vec<Value> = team.members.iter().map(|uid| {
+                let user = user_store.users.values().find(|u| u.id == *uid);
+                json!({
+                    "id": uid,
+                    "email": user.map(|u| u.email.as_str()).unwrap_or("unknown"),
+                    "name": user.and_then(|u| u.name.as_deref()),
+                })
+            }).collect();
+
+            (StatusCode::OK, Json(json!({
+                "team": {
+                    "id": team.id,
+                    "name": team.name,
+                    "created_by": team.created_by,
+                    "members": members,
+                    "created_at": team.created_at,
+                }
+            })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
+#[derive(Deserialize)]
+struct AddMemberRequest {
+    email: String,
+}
+
+async fn add_member(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path(id): Path<String>,
+    Json(body): Json<AddMemberRequest>,
+) -> (StatusCode, Json<Value>) {
+    let email = body.email.trim().to_lowercase();
+
+    let user_store = state.user_store.read().await;
+    let target_user = user_store.find_by_email(&email);
+    let target_id = match target_user {
+        Some(u) => u.id.clone(),
+        None => return (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
+    };
+    drop(user_store);
+
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            if team.members.contains(&target_id) {
+                return (StatusCode::CONFLICT, Json(json!({ "error": "Already a member" })));
+            }
+            team.members.push(target_id);
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
+async fn remove_member(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path((id, user_id)): Path<(String, String)>,
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
+// ── Tests ────────────────────────────────────────────────────
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn team_store_roundtrip() {
+        let tmp = tempfile::TempDir::new().unwrap();
+        let dir = tmp.path().to_path_buf();
+
+        let mut store = TeamStore { teams: HashMap::new() };
+        store.teams.insert("t1".into(), Team {
+            id: "t1".into(),
+            name: "Alpha".into(),
+            created_by: "u1".into(),
+            members: vec!["u1".into(), "u2".into()],
+            created_at: "2026-01-01".into(),
+        });
+        store.save(&dir).unwrap();
+
+        let loaded = TeamStore::load(&dir);
+        assert_eq!(loaded.teams.len(), 1);
+        assert_eq!(loaded.teams["t1"].name, "Alpha");
+        assert_eq!(loaded.teams["t1"].members.len(), 2);
+    }
+
+    #[test]
+    fn teams_for_user_filters_correctly() {
+        let mut store = TeamStore { teams: HashMap::new() };
+        store.teams.insert("t1".into(), Team {
+            id: "t1".into(), name: "A".into(), created_by: "u1".into(),
+            members: vec!["u1".into(), "u2".into()], created_at: "".into(),
+        });
+        store.teams.insert("t2".into(), Team {
+            id: "t2".into(), name: "B".into(), created_by: "u3".into(),
+            members: vec!["u3".into()], created_at: "".into(),
+        });
+
+        assert_eq!(store.teams_for_user("u1").len(), 1);
+        assert_eq!(store.teams_for_user("u2").len(), 1);
+        assert_eq!(store.teams_for_user("u3").len(), 1);
+        assert_eq!(store.teams_for_user("u99").len(), 0);
+    }
+}
--- a/cthulu-studio/src/api/client.ts
+++ b/cthulu-studio/src/api/client.ts
@@ -37,6 +37,39 @@ export function getServerUrl(): string {
   return getBaseUrl();
 }
 
+// ── Auth token injection ─────────────────────────────────────
+
+let getTokenFn: (() => Promise<string | null>) | null = null;
+
+/** Called by AuthGate to wire token getter into all API requests. */
+export function setAuthTokenGetter(fn: (() => Promise<string | null>) | null) {
+  getTokenFn = fn;
+}
+
+/** Get the current auth token (if available). Used by SSE EventSource callers. */
+export async function getAuthToken(): Promise<string | null> {
+  return getTokenFn ? getTokenFn() : null;
+}
+
+/** Get auth token synchronously from localStorage (for EventSource URLs). */
+export function getAuthTokenSync(): string | null {
+  return localStorage.getItem("cthulu_auth_token");
+}
+
+/** Append ?token= to a URL for EventSource connections (which can't set headers). */
+export function withAuthToken(url: string): string {
+  const token = getAuthTokenSync();
+  if (!token) return url;
+  const sep = url.includes("?") ? "&" : "?";
+  return `${url}${sep}token=${encodeURIComponent(token)}`;
+}
+
+/** Get auth headers for fetch-based streaming (non-apiFetch callers). */
+export async function getAuthHeaders(): Promise<Record<string, string>> {
+  const token = await getAuthToken();
+  return token ? { Authorization: `Bearer ${token}` } : {};
+}
+
 async function apiFetch<T>(
   path: string,
   options: RequestInit = {}
@@ -47,11 +80,21 @@ async function apiFetch<T>(
   log("http", `${method} ${path}`);
   const start = performance.now();
 
+  // Attach auth token if available
+  const authHeaders: Record<string, string> = {};
+  if (getTokenFn) {
+    const token = await getTokenFn();
+    if (token) {
+      authHeaders["Authorization"] = `Bearer ${token}`;
+    }
+  }
+
   try {
     const res = await fetch(url, {
       ...options,
       headers: {
         "Content-Type": "application/json",
+        ...authHeaders,
         ...options.headers,
       },
     });
@@ -59,6 +102,11 @@ async function apiFetch<T>(
     const elapsed = Math.round(performance.now() - start);
 
     if (!res.ok) {
+      // Auto-logout on 401 (expired/invalid token)
+      if (res.status === 401 && getTokenFn) {
+        localStorage.removeItem("cthulu_auth_token");
+        window.location.reload();
+      }
       const body = await res.text();
       log("error", `${method} ${path} -> ${res.status} (${elapsed}ms)`, body);
       throw new Error(`API error ${res.status}: ${body}`);
@@ -326,7 +374,7 @@ export function streamSessionLog(
   onLine: (line: string) => void,
   onDone: () => void
 ): () => void {
-  const url = `${getBaseUrl()}/api/agents/${agentId}/sessions/${sessionId}/stream`;
+  const url = withAuthToken(`${getBaseUrl()}/api/agents/${agentId}/sessions/${sessionId}/stream`);
   const source = new EventSource(url);
 
   source.addEventListener("line", (e: MessageEvent) => {
@@ -586,7 +634,7 @@ export interface ResourceChangeEvent {
 export function subscribeToChanges(
   onEvent: (event: ResourceChangeEvent) => void
 ): () => void {
-  const url = `${getBaseUrl()}/api/changes`;
+  const url = withAuthToken(`${getBaseUrl()}/api/changes`);
   const source = new EventSource(url);
 
   const handler = (e: MessageEvent) => {
@@ -706,4 +754,68 @@ export async function getDashboardSummary(
     method: "POST",
     body: JSON.stringify({ channels }),
   });
+}
+
+// ── Profile ──────────────────────────────────────────────────
+
+export interface UserProfile {
+  id: string;
+  email: string;
+  name: string | null;
+  avatar_url: string | null;
+  created_at: string;
+}
+
+export async function getProfile(): Promise<UserProfile> {
+  return apiFetch<UserProfile>("/auth/me");
+}
+
+export async function updateProfile(data: { name?: string; avatar_url?: string }): Promise<UserProfile> {
+  return apiFetch<UserProfile>("/auth/me", { method: "PUT", body: JSON.stringify(data) });
+}
+
+export interface UserSearchResult {
+  id: string;
+  email: string;
+  name: string | null;
+}
+
+export async function searchUsers(query: string): Promise<{ users: UserSearchResult[] }> {
+  return apiFetch<{ users: UserSearchResult[] }>(`/auth/users/search?q=${encodeURIComponent(query)}`);
+}
+
+// ── Teams ────────────────────────────────────────────────────
+
+export interface TeamMember {
+  id: string;
+  email: string;
+  name: string | null;
+}
+
+export interface Team {
+  id: string;
+  name: string;
+  created_by: string;
+  members: TeamMember[] | string[];
+  created_at: string;
+}
+
+export async function listTeams(): Promise<{ teams: Team[] }> {
+  return apiFetch<{ teams: Team[] }>("/teams");
+}
+
+export async function createTeam(name: string): Promise<{ team: Team }> {
+  return apiFetch<{ team: Team }>("/teams", { method: "POST", body: JSON.stringify({ name }) });
+}
+
+export async function getTeam(id: string): Promise<{ team: Team }> {
+  return apiFetch<{ team: Team }>(`/teams/${id}`);
+}
+
+export async function addTeamMember(teamId: string, email: string): Promise<{ ok: boolean }> {
+  return apiFetch<{ ok: boolean }>(`/teams/${teamId}/members`, { method: "POST", body: JSON.stringify({ email }) });
+}
+
+export async function removeTeamMember(teamId: string, userId: string): Promise<{ ok: boolean }> {
+  return apiFetch<{ ok: boolean }>(`/teams/${teamId}/members/${userId}`, { method: "DELETE" });
 }
--- a/cthulu-studio/src/components/AuthGate.tsx
+++ b/cthulu-studio/src/components/AuthGate.tsx
@@ -1,0 +1,176 @@
+import { useState, useCallback, createContext, useContext } from "react";
+import { setAuthTokenGetter, getServerUrl } from "../api/client";
+
+// ── Auth context so any component can call logout ────────────
+
+interface AuthContextValue {
+  logout: () => void;
+  userEmail: string | null;
+}
+
+const AuthContext = createContext<AuthContextValue>({
+  logout: () => {},
+  userEmail: null,
+});
+
+/** Hook for any component to access logout + user info. */
+export function useAuth() {
+  return useContext(AuthContext);
+}
+
+interface AuthGateProps {
+  children: React.ReactNode;
+}
+
+/** Decode base64url to string (handles missing padding). */
+function b64urlDecode(str: string): string {
+  const padded = str.replace(/-/g, "+").replace(/_/g, "/") + "=".repeat((4 - (str.length % 4)) % 4);
+  return atob(padded);
+}
+
+/** Parse JWT payload and return claims, or null if unparseable. */
+function parseToken(token: string): { email?: string; exp?: number } | null {
+  try {
+    const payload = token.split(".")[1];
+    return JSON.parse(b64urlDecode(payload));
+  } catch {
+    return null;
+  }
+}
+
+/** Check if token is valid: parseable, has email, not expired. */
+function isTokenValid(token: string): boolean {
+  const claims = parseToken(token);
+  if (!claims?.email) return false;
+  if (claims.exp && claims.exp * 1000 < Date.now()) return false;
+  return true;
+}
+
+/**
+ * Self-hosted auth gate. Shows login/signup when no token,
+ * renders children when authenticated.
+ * When VITE_AUTH_ENABLED is not "true", renders children directly (dev mode).
+ */
+export default function AuthGate({ children }: AuthGateProps) {
+  const authEnabled = import.meta.env.VITE_AUTH_ENABLED === "true";
+
+  // Initialize token AND token getter synchronously to avoid race condition
+  // where children's effects fire API calls before the getter is wired up.
+  const [token, setToken] = useState<string | null>(() => {
+    const t = localStorage.getItem("cthulu_auth_token");
+    if (t && isTokenValid(t)) {
+      setAuthTokenGetter(() => Promise.resolve(t));
+      return t;
+    }
+    // Clear invalid/expired token
+    if (t) localStorage.removeItem("cthulu_auth_token");
+    setAuthTokenGetter(null);
+    return null;
+  });
+
+  const logout = useCallback(() => {
+    localStorage.removeItem("cthulu_auth_token");
+    setAuthTokenGetter(null);
+    setToken(null);
+  }, []);
+
+  if (!authEnabled) return <>{children}</>;
+
+  if (token) {
+    const claims = parseToken(token);
+    return (
+      <AuthContext.Provider value={{ logout, userEmail: claims?.email ?? null }}>
+        {children}
+      </AuthContext.Provider>
+    );
+  }
+
+  return <AuthForm onSuccess={(t) => {
+    localStorage.setItem("cthulu_auth_token", t);
+    setAuthTokenGetter(() => Promise.resolve(t));
+    setToken(t);
+  }} />;
+}
+
+function AuthForm({ onSuccess }: { onSuccess: (token: string) => void }) {
+  const [mode, setMode] = useState<"login" | "signup">("login");
+  const [email, setEmail] = useState("");
+  const [password, setPassword] = useState("");
+  const [error, setError] = useState("");
+  const [loading, setLoading] = useState(false);
+
+  const handleSubmit = async (e: React.FormEvent) => {
+    e.preventDefault();
+    setError("");
+    setLoading(true);
+
+    try {
+      const endpoint = mode === "signup" ? "/api/auth/signup" : "/api/auth/login";
+      const res = await fetch(`${getServerUrl()}${endpoint}`, {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ email, password }),
+      });
+
+      const data = await res.json();
+
+      if (!res.ok) {
+        setError(data.error || `Failed (${res.status})`);
+        return;
+      }
+
+      onSuccess(data.token);
+    } catch (err) {
+      setError(err instanceof Error ? err.message : "Network error");
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  return (
+    <div className="auth-container">
+      <div className="auth-card">
+        <h1 className="auth-title">Cthulu</h1>
+        <p className="auth-subtitle">
+          {mode === "login" ? "Sign in to your account" : "Create a new account"}
+        </p>
+
+        <form onSubmit={handleSubmit} className="auth-form">
+          <input
+            type="email"
+            placeholder="Email"
+            value={email}
+            onChange={(e) => setEmail(e.target.value)}
+            className="auth-input"
+            required
+            autoFocus
+          />
+          <input
+            type="password"
+            placeholder="Password (8+ characters)"
+            value={password}
+            onChange={(e) => setPassword(e.target.value)}
+            className="auth-input"
+            required
+            minLength={8}
+          />
+
+          {error && <p className="auth-error">{error}</p>}
+
+          <button type="submit" className="auth-button" disabled={loading}>
+            {loading ? "..." : mode === "login" ? "Sign In" : "Sign Up"}
+          </button>
+        </form>
+
+        <button
+          className="auth-toggle"
+          onClick={() => { setMode(mode === "login" ? "signup" : "login"); setError(""); }}
+        >
+          {mode === "login"
+            ? "Don't have an account? Sign up"
+            : "Already have an account? Sign in"}
+        </button>
+      </div>
+    </div>
+  );
+}
--- a/cthulu-studio/src/components/TopBar.tsx
+++ b/cthulu-studio/src/components/TopBar.tsx
@@ -1,4 +1,5 @@
 import { useState, useEffect, useRef, useCallback } from "react";
+import { useAuth } from "./AuthGate";
 import * as api from "../api/client";
 import { log } from "../api/logger";
 import { Button } from "@/components/ui/button";
@@ -249,6 +250,8 @@ export default function TopBar({
       <Button variant="ghost" size="sm" onClick={onSettingsClick}>
         Settings
       </Button>
+
+      <ProfileMenu />
     </div>
   );
 }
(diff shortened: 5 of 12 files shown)


saltyskip:

_Reviewing PR #121 — this may take a few minutes._

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,517 @@
...
+                Json(json!({ "error": "User already exists" })),
+            );
+        }
+    }
+
+    let password_hash = match bcrypt::hash(&body.password, 12) {
+        Ok(h) => h,
+        Err(e) => {
+            return (
+                StatusCode::INTERNAL_SERVER_ERROR,
```
saltyskip, line 298: **Bug: TOCTOU race condition in signup.** The read lock is dropped at line 257, then a separate write lock is taken at line 289. Between those two points, a concurrent signup request with the same email can pass the existence check and insert a duplicate user.

Fix: use a single write lock for the check-and-insert:
```rust
}
store.insert(user.clone());
```

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,517 @@
...
+    match std::fs::read_to_string(&path) {
+        Ok(secret) if !secret.trim().is_empty() => secret.trim().to_string(),
+        _ => {
+            let secret = uuid::Uuid::new_v4().to_string();
+            let _ = std::fs::write(&path, &secret);
+            secret
+        }
+    }
+}
+
```
saltyskip, line 121: **Weak JWT secret.** `uuid::Uuid::new_v4()` gives 122 bits of randomness in a predictable hex-dash format. For a signing secret, use a proper cryptographic random source:
```rust
use rand::RngCore;
let mut bytes = [0u8; 32];
rand::thread_rng().fill_bytes(&mut bytes);
let secret = hex::encode(bytes);
```
(You already have `rand` as a transitive dep via `bcrypt`.)

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,517 @@
...
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("users.json");
+        let json = serde_json::to_string_pretty(&self.users)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn find_by_email(&self, email: &str) -> Option<&StoredUser> {
+        self.users.get(email)
+    }
+
```
saltyskip, line 79: **Non-atomic file write.** Both `UserStore::save` and `TeamStore::save` use `std::fs::write` directly, which can corrupt data if the process crashes mid-write. Per the project's own rule #9 (atomic YAML persistence), use temp-file + rename:
```rust
std::fs::write(&tmp, json)?;
std::fs::rename(tmp, path)
```

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,517 @@
...
+    }
+}
+
+pub fn auth_enabled() -> bool {
+    std::env::var("AUTH_ENABLED")
+        .map(|v| v == "true")
+        .unwrap_or(false)
+}
+
+impl FromRequestParts<AppState> for AuthUser {
```
saltyskip, line 180: **Config duplication.** `auth_enabled()` reads `AUTH_ENABLED` directly from the environment, but the PR also adds `auth_enabled` to the `Config` struct (in `config.rs`). The `Config` field is never used — `clerk_auth` and `user_context` both call this function instead. Pick one source of truth: either use `Config.auth_enabled` (pass it through `AppState`) or remove the field from `Config`.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,517 @@
...
+    axum::extract::State(state): axum::extract::State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.user_store.read().await;
+    let user = store.users.values().find(|u| u.id == auth.user_id);
+    match user {
+        Some(u) => (StatusCode::OK, Json(json!({
+            "id": u.id, "email": u.email, "name": u.name,
+            "avatar_url": u.avatar_url, "created_at": u.created_at,
+        }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
```
saltyskip, line 407: **O(n) linear scan on every profile request.** The `UserStore` is keyed by email, but `get_profile` and `update_profile` both do `.values().find(|u| u.id == auth.user_id)` — a full linear scan. As the user count grows this gets expensive on every authenticated request.

Consider adding a secondary `HashMap<String, String>` (id → email) index, or changing the primary key to user ID.

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,152 @@
...
+  if (!authEnabled) return <>{children}</>;
+
+  if (token) {
+    setAuthTokenGetter(() => Promise.resolve(token));
+    return (
+      <AuthContext.Provider value={{ logout, userEmail: emailFromToken(token) }}>
+        {children}
+      </AuthContext.Provider>
+    );
+  }
```
saltyskip, line 86: **Side effect during render.** `setAuthTokenGetter(() => Promise.resolve(token))` is called unconditionally in the render body (not in a `useEffect` or event handler). This mutates module-level state on every render, which React Strict Mode will call twice. Move this into a `useEffect`:
```tsx
useEffect(() => {
}, [token]);
```

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+  const [teamMembers, setTeamMembers] = useState<Record<string, { id: string; email: string; name: string | null }[]>>({});
+  const [memberError, setMemberError] = useState("");
+  const menuRef = useRef<HTMLDivElement>(null);
+
+  if (!userEmail) return null;
+
+  const loadData = useCallback(async () => {
+    try {
+      const [p, t] = await Promise.all([api.getProfile(), api.listTeams()]);
+      setProfile({ name: p.name, email: p.email });
```
saltyskip, line 310: **`useCallback` with empty deps but calls `setProfile`/`setTeams`.** `loadData` closes over `setProfile` and `setTeams` which are stable (React state setters), so this works — but the empty deps array `[]` is misleading if `api.getProfile` or `api.listTeams` ever took parameters from component state. Minor, but worth noting.

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+  const handleExpandTeam = async (teamId: string) => {
+    if (expandedTeam === teamId) { setExpandedTeam(null); return; }
+    setExpandedTeam(teamId);
+    setMemberError("");
+    try { await loadTeamMembers(teamId); } catch { /* ignore */ }
+  };
+
+  const handleRemoveMember = async (teamId: string, userId: string) => {
+    try {
+      await api.removeTeamMember(teamId, userId);
```
saltyskip, line 349: **`searching` state can get stuck.** When the user types fast, `setSearching(true)` fires on each keystroke but the previous `setTimeout` is cleared before its `finally { setSearching(false) }` runs. This means `searching` stays `true` until the final debounced call completes. Consider moving `setSearching(true)` inside the timeout callback, right before the `await`.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
+        self.teams.values()
+            .filter(|t| t.members.contains(&user_id.to_string()))
+            .collect()
```
saltyskip, line 51: Same non-atomic write issue as `UserStore::save` — use temp-file + rename to prevent data corruption on crash.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
```
saltyskip, line 190: **User can remove themselves from a team (and auto-delete it).** If the last member removes themselves (line 191-193), the team is silently deleted. This might be intentional, but there's no confirmation or protection — consider at least logging this, or returning a response indicating the team was dissolved.

saltyskip (changes requested):
## 🤖 Cthulu Review

Good feature set — auth, profile, teams, and the UI all hang together well. Tests are solid. A few issues need addressing before merge:

### Must fix
1. **TOCTOU race in signup** — read lock dropped before write lock taken, allowing duplicate users under concurrent requests. Use a single write lock for check-and-insert.
2. **Side effect during render in AuthGate** — `setAuthTokenGetter` is called in the render body, not in a `useEffect`. Will fire twice in React Strict Mode and is a correctness issue.
3. **Non-atomic file persistence** — both `UserStore::save` and `TeamStore::save` use `std::fs::write` directly. A crash mid-write corrupts data. Use temp-file + rename per project rule #9.

### Should fix
4. **Weak JWT secret** — UUID v4 is not ideal for a signing secret. Use `rand::thread_rng().fill_bytes()` for 256 bits of crypto randomness.
5. **Duplicate `auth_enabled` config** — `clerk_auth::auth_enabled()` reads from env directly, making the `Config.auth_enabled` field dead code. Pick one source of truth.
6. **O(n) user lookup by ID** — `get_profile`/`update_profile` do linear scans over the email-keyed hashmap. Add an ID→email index or re-key by ID.
7. **`searching` state stuck true** — in `UserSearchInput`, `setSearching(true)` fires on keystroke but the cleared timeout never runs `setSearching(false)`.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #121 (previous HEAD: `cdd8649`, new HEAD: `a06b3f1`)_

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+
+function ProfileMenu() {
+  const { logout, userEmail } = useAuth();
+  const [open, setOpen] = useState(false);
+  const [profile, setProfile] = useState<{ name: string | null; email: string } | null>(null);
+  const [teams, setTeams] = useState<{ id: string; name: string }[]>([]);
+  const [editingName, setEditingName] = useState(false);
+  const [nameInput, setNameInput] = useState("");
+  const [newTeamName, setNewTeamName] = useState("");
+  const [expandedTeam, setExpandedTeam] = useState<string | null>(null);
```
saltyskip, line 300: **Bug (crash): React hooks violation.** `useCallback` (line 302) and `useEffect` (lines 310, 315) are called *after* the early return on line 300. React hooks must be called unconditionally in the same order on every render. If `userEmail` starts as `null` (e.g., during auth check) and then becomes non-null, the hook count changes and React will throw:

> Rendered more hooks than during the previous render.

Fix: move the early return *after* all hook declarations, or guard the return inside the JSX:

```tsx
function ProfileMenu() {
  const { logout, userEmail } = useAuth();
  const [open, setOpen] = useState(false);
  // ... all useState/useRef ...

  const loadData = useCallback(async () => { ... }, []);

  useEffect(() => { if (open) loadData(); }, [open, loadData]);
  useEffect(() => { ... click-outside ... }, [open]);

  if (!userEmail) return null;  // <-- move here, after all hooks
  // ... rest of component
}
```

cthulu-backend/api/teams.rs
saltyskip, line 51: **Inconsistency: `TeamStore::save` is not atomic.** `UserStore::save` correctly uses temp-file + rename for crash safety (per project rule #9), but `TeamStore::save` uses a direct `std::fs::write`. If the process crashes mid-write, `teams.json` will be corrupted.

Fix:
```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,523 @@
...
+    }
+}
+
+#[derive(Deserialize)]
+struct UpdateProfileRequest {
+    name: Option<String>,
+    avatar_url: Option<String>,
```
saltyskip, line 414: **Performance: O(n) linear scan on every profile request.** The `UserStore` HashMap is keyed by email, but `get_profile` and `update_profile` look up by `user_id` via `.values().find()`. With many users this becomes a bottleneck since it runs on every authenticated request to `/auth/me`.

Consider adding a secondary index `id_to_email: HashMap<String, String>` to `UserStore` for O(1) lookups by user ID, or re-keying the primary map by ID.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,523 @@
...
+    if let Some(query) = parts.uri.query() {
+        for pair in query.split('&') {
+            if let Some(token) = pair.strip_prefix("token=") {
+                return Some(token.to_string());
+            }
+        }
+    }
+
+    None
+}
```
saltyskip, line 243: **Security concern: token in query string.** Accepting JWT via `?token=` means tokens leak into server access logs, browser history, HTTP Referer headers, and proxy logs. If this is only needed for SSE/EventSource (which can't set headers), consider limiting this to the SSE endpoint specifically rather than accepting it globally on all routes.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
```
saltyskip, line 189: **Authorization: any member can remove any other member, including the creator.** There's no check on whether the requesting user should have permission to remove members. A team member can remove the team creator (or themselves, causing the team to auto-delete if `members` becomes empty).

Consider at minimum preventing removal of the `created_by` user, or adding a `role` field to control who can manage membership.

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+}
+
+function UserSearchInput({ onSelect }: { onSelect: (userId: string, email: string) => void }) {
+  const [query, setQuery] = useState("");
+  const [results, setResults] = useState<api.UserSearchResult[]>([]);
+  const [searching, setSearching] = useState(false);
+  const debounceRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  const handleChange = (value: string) => {
+    setQuery(value);
```
saltyskip, line 441: **Minor: debounce timer not cleaned up on unmount.** If `UserSearchInput` unmounts while a debounced timeout is pending, the `setTimeout` callback will fire and call `setResults`/`setSearching` on an unmounted component. Add a cleanup:

```tsx
useEffect(() => {
  return () => { if (debounceRef.current) clearTimeout(debounceRef.current); };
}, []);
```

saltyskip (changes requested):
## 🤖 Cthulu Review

**Request changes** — one crash bug, one data-safety issue.

### Blocking

1. **React hooks violation in `ProfileMenu`** — `useCallback`/`useEffect` are called after an early return (`if (!userEmail) return null`). This violates React's rules of hooks and will crash at runtime when `userEmail` transitions from null to non-null. Move the early return after all hook declarations.

2. **`TeamStore::save` is not atomic** — Uses direct `std::fs::write` instead of temp-file + rename, unlike `UserStore::save`. A crash mid-write corrupts `teams.json`. This violates project rule #9.

### Non-blocking

3. **O(n) user lookup by ID** — `get_profile` and `update_profile` do `values().find()` on an email-keyed HashMap. Add a secondary index or re-key by ID.

4. **Token accepted in query string globally** — JWT in `?token=` leaks to logs/history/referrers. Consider restricting to SSE endpoints only.

5. **No authorization on member removal** — Any team member can remove any other member, including the creator. Consider role-based checks or at minimum protect the creator.

6. **Debounce timer not cleaned up on unmount** in `UserSearchInput`.

Overall the auth implementation is solid — good bcrypt usage with `spawn_blocking`, atomic user persistence, proper TOCTOU prevention on signup, and clean dev-mode bypass. The tests are thorough. Just need to fix the hooks violation and atomic save before merging.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #121 (previous HEAD: `a06b3f1`, new HEAD: `ec6fabe`)_

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+    if (open) loadData();
+  }, [open, loadData]);
+
+  useEffect(() => {
+    if (!open) return;
+    const handler = (e: MouseEvent) => {
+      if (menuRef.current && !menuRef.current.contains(e.target as Node)) setOpen(false);
+    };
+    document.addEventListener("mousedown", handler);
+    return () => document.removeEventListener("mousedown", handler);
```
saltyskip, line 325: **Bug (will crash React):** `useCallback` and the two `useEffect` calls below are placed _after_ this conditional early return. React requires hooks to be called in the same order on every render — if `userEmail` toggles between truthy/falsy across renders, the hook count changes and React throws.

Move the early return to **after** all hook declarations (after the two `useEffect` blocks), or restructure so the component always calls the same hooks.

```tsx
// Move ALL hooks above this line:
const loadData = useCallback(...);
useEffect(...);
useEffect(...);

if (!userEmail) return null; // safe here — hooks already called
```

cthulu-backend/api/teams.rs
saltyskip, line 51: **Not atomic:** `TeamStore::save` writes directly via `std::fs::write`, unlike `UserStore::save` which uses temp-file + rename for crash safety. This violates the project's rule #9 (atomic persistence). A crash mid-write corrupts `teams.json`.

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,530 @@
...
+
+    if let Some(query) = parts.uri.query() {
+        for pair in query.split('&') {
+            if let Some(token) = pair.strip_prefix("token=") {
+                return Some(token.to_string());
+            }
+        }
+    }
+
+    None
```
saltyskip, line 242: **Bug:** The token value extracted from the query string is not URL-decoded. If a JWT contains `+`, `%`, or `&` characters (base64url tokens can contain `+`), this will silently truncate or corrupt the token. Consider using `urlencoding::decode(token)` or a proper query parser, or just drop query-param auth since the Bearer header is the primary path.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,530 @@
...
+    axum::extract::State(state): axum::extract::State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.user_store.read().await;
+    let user = store.users.values().find(|u| u.id == auth.user_id);
+    match user {
+        Some(u) => (StatusCode::OK, Json(json!({
+            "id": u.id, "email": u.email, "name": u.name,
+            "avatar_url": u.avatar_url, "created_at": u.created_at,
+        }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
```
saltyskip, line 407: **Performance:** `get_profile` does a linear scan over all users (`.values().find(|u| u.id == auth.user_id)`) because `UserStore` is keyed by email, not by user ID. Same issue in `update_profile`. With many users this becomes slow.

Consider adding a second index (`HashMap<String, String>` mapping `id -> email`) to `UserStore`, or re-key the primary map by ID and add an email index for login lookups.

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+      <input className="auth-input" placeholder="Add member (search name or email)"
+        value={query} onChange={(e) => handleChange(e.target.value)} />
+      {(results.length > 0 || searching) && (
+        <div className="user-search-results">
+          {searching && <div className="user-search-item user-search-loading">Searching...</div>}
+          {results.map((u) => (
+            <button key={u.id} className="user-search-item"
+              onClick={() => { onSelect(u.id, u.email); setQuery(""); setResults([]); }}>
+              <span className="user-search-name">{u.name || u.email.split("@")[0]}</span>
+              <span className="user-search-email">{u.email}</span>
```
saltyskip, line 466: **Race condition in debounce:** `setSearching(true)` is called synchronously, but the API call runs inside `setTimeout`. If the user types another character before the 300ms fires, the old timeout is cleared but `searching` remains `true` with no pending request to set it `false`. Also, there's no cancellation of in-flight requests — a slow response from an earlier keystroke can overwrite results from a later, faster one.

Consider using an `AbortController` or a request-id counter to discard stale responses.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,530 @@
...
+    };
+
+    store.insert(user.clone());
+    let _ = store.save(&state.data_dir);
+    drop(store);
+
+    (
+        StatusCode::CREATED,
+        Json(json!({
+            "token": token,
```
saltyskip, line 341: Nit: `let _ = store.save(...)` silently swallows save errors here and in several other places (teams.rs too). If the disk is full or permissions fail, the in-memory state diverges from disk with no signal. Consider at minimum logging the error:

```rust
    tracing::error!("Failed to persist user store: {e}");
}
```

cthulu-backend/api/teams.rs
saltyskip, line 190: **Design consideration:** Any team member can remove any other member (including the creator), and if the last member removes themselves the team is auto-deleted. This might be intentional for a flat/no-roles model, but it means any member can effectively destroy the team. Worth confirming this is the desired behavior, or at minimum preventing removal of the `created_by` user.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Request changes** — one crash-level bug and a few things to fix before merging.

### Blocking

1. **React hooks violation in `ProfileMenu`** — `useCallback` and two `useEffect` calls are placed after a conditional early return (`if (!userEmail) return null`). This will crash React when `userEmail` changes between renders. All hooks must be moved above the early return.

2. **`TeamStore::save()` is not atomic** — Uses direct `std::fs::write` instead of temp-file + rename like `UserStore::save()`. Violates project rule #9 and risks data corruption on crash.

### Should fix

3. **Query param token not URL-decoded** — `extract_token` reads the raw query string without decoding, which can silently truncate or corrupt tokens containing `&` or `%`-encoded chars.

4. **`UserSearchInput` debounce race condition** — `setSearching(true)` fires synchronously but the timeout callback may be cleared without resetting it. No cancellation of stale in-flight requests either.

5. **Save errors silently swallowed** — `let _ = store.save(...)` in both user and team stores discards I/O errors. At minimum log them.

### Non-blocking observations

- `get_profile` / `update_profile` do O(n) linear scans because `UserStore` is keyed by email but lookups are by user ID. Consider a secondary index.
- Any team member can remove any other member (including the creator) and auto-delete the team. Confirm this is intentional for the flat-roles model.

Overall the auth system is well-structured — good use of `spawn_blocking` for bcrypt, proper TOCTOU prevention on signup, and clean dev-mode bypass. Just needs the hooks fix and atomic save before merging.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #121 (previous HEAD: `ec6fabe`, new HEAD: `6cffd28`)_

cthulu-studio/src/components/TopBar.tsx
saltyskip, line 300: **Bug: Hooks called after early return — violates React Rules of Hooks.**

`useCallback` on line 302 and `useEffect` on lines 310/314 are called *after* `if (!userEmail) return null;` on line 300. React requires hooks to be called in the same order on every render. If `userEmail` transitions from null to non-null (e.g., after login), React will see a different number of hooks and throw a runtime error.

Fix: move the early return *below* all hook calls, or restructure so hooks are unconditional:

```tsx
const loadData = useCallback(async () => { ... }, []);
useEffect(() => { if (open) loadData(); }, [open, loadData]);
useEffect(() => { ... }, [open]);

if (!userEmail) return null;
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
+        self.teams.values()
+            .filter(|t| t.members.contains(&user_id.to_string()))
+            .collect()
```
saltyskip, line 51: **Non-atomic write — data loss risk on crash.**

`TeamStore::save()` uses direct `std::fs::write(path, json)`, while `UserStore::save()` (in `local_auth.rs`) correctly uses temp file + rename for crash-safe atomic writes. If the process crashes mid-write here, `teams.json` will be corrupted (truncated/partial).

Fix: use the same pattern as `UserStore::save()`:
```rust
std::fs::write(&tmp, json)?;
std::fs::rename(&tmp, &path)
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
```
saltyskip, line 190: **Any team member can remove any other member (including the creator).**

The only authorization check is `team.members.contains(&auth.user_id)` — any member can remove any other member. Additionally, if a member removes themselves and they're the last one, the team is silently deleted (line 191-193). Consider at minimum preventing removal of `team.created_by`, or at least documenting this as intentional flat-access design.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,554 @@
...
+    axum::extract::State(state): axum::extract::State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.user_store.read().await;
+    let user = store.users.values().find(|u| u.id == auth.user_id);
+    match user {
+        Some(u) => (StatusCode::OK, Json(json!({
+            "id": u.id, "email": u.email, "name": u.name,
+            "avatar_url": u.avatar_url, "created_at": u.created_at,
+        }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
```
saltyskip, line 407: **O(n) scan on every profile fetch.** `UserStore` is keyed by email, but `get_profile` scans all values to find by `user_id`. Same issue in `update_profile`. For a small user base this is fine, but consider adding a secondary `id -> email` index if this store is expected to grow, or at minimum add a `find_by_id` helper to centralize the scan.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,554 @@
...
+        .route("/auth/users/search", get(search_users))
+}
+
+async fn signup(
+    axum::extract::State(state): axum::extract::State<AppState>,
+    Json(body): Json<SignupRequest>,
+) -> (StatusCode, Json<Value>) {
+    let email = body.email.trim().to_lowercase();
+    if email.is_empty() || !email.contains('@') {
+        return (
```
saltyskip, line 274: **No rate limiting on signup/login.** These endpoints are open (no auth) and do bcrypt hashing (CPU-intensive). Without rate limiting, an attacker can exhaust server CPU via repeated signup requests (bcrypt cost=12 is ~250ms per hash). Consider adding basic rate limiting middleware or at least per-IP throttling for auth endpoints.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Must fix before merge:**

1. **React Rules of Hooks violation in `ProfileMenu`** — `useCallback` and `useEffect` are called after an early `return null` guard. This will crash at runtime when `userEmail` changes between renders. Move hooks above the early return.

2. **`TeamStore::save()` is not atomic** — Uses direct `std::fs::write` while `UserStore::save()` correctly uses temp file + rename. A crash during write will corrupt `teams.json`. Use the same atomic pattern.

**Worth considering:**

- Any team member can remove any other member including the creator, and removing the last member silently deletes the team. If this flat-access model is intentional, a comment would help.
- `get_profile` and `update_profile` do O(n) scans since the user store is keyed by email but lookups are by user ID.
- No rate limiting on open signup/login endpoints — bcrypt at cost=12 makes these CPU-expensive and abuse-friendly.

Overall the auth/teams implementation is solid — JWT handling, bcrypt offloading to `spawn_blocking`, atomic user store persistence, TOCTOU prevention on signup, and the `AuthUser` extractor design are all well done. Just need to fix the two blocking issues above.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #121 (previous HEAD: `6cffd28`, new HEAD: `d35c2cb`)_

cthulu-studio/src/components/TopBar.tsx
saltyskip, line 325: **Bug: Rules of Hooks violation.** `useCallback` and `useEffect` are called after this conditional early return. React hooks must be called unconditionally and in the same order on every render. When `userEmail` transitions between `null` and a value, React will throw.

Fix: move the early return below all hook calls, or restructure to avoid the conditional return:

```tsx
const loadData = useCallback(async () => {
  // ...
}, []);

useEffect(() => {
  if (open) loadData();
}, [open, loadData]);

useEffect(() => {
  if (!open) return;
  // click-outside handler...
}, [open]);

if (!userEmail) return null; // <-- move here, after all hooks
```

cthulu-backend/api/teams.rs
saltyskip, line 51: **Non-atomic write.** `UserStore::save` in `local_auth.rs` correctly uses temp file + rename for crash safety, but `TeamStore::save` does a direct `std::fs::write`. A crash mid-write would corrupt `teams.json`. Use the same pattern:

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-studio/src/components/TopBar.tsx
saltyskip, line 349: No error handling on `updateProfile`. If the API call fails, the user gets no feedback and the local state will be inconsistent (name reverts on next open). Wrap in try/catch and surface the error.

cthulu-backend/api/local_auth.rs
saltyskip, line 407: Minor: `store.users` is keyed by email but this does a linear scan by `user_id`. Same in `update_profile`. Fine for small stores, but consider adding a secondary `id -> email` index if you expect this to grow, or key the HashMap by `id` instead.

cthulu-backend/api/teams.rs
saltyskip, line 190: Any member can remove any other member (including the team creator), and if the last member removes themselves, the team is auto-deleted. This means any member can effectively delete the team by removing everyone. Is this the intended access model? Consider restricting removal to `created_by` or at least preventing removal of the creator.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Must fix before merge:**

1. **Rules of Hooks violation in `ProfileMenu`** (`TopBar.tsx:325`): `useCallback` and `useEffect` are called after a conditional `return null`. This will crash React when `userEmail` transitions between null and a value. Move the early return below all hook calls.

2. **Non-atomic `TeamStore::save`** (`teams.rs:51`): Uses direct `std::fs::write`, unlike `UserStore::save` which correctly uses temp file + rename. A crash mid-write corrupts `teams.json`.

**Should fix:**

3. **No error handling on `handleSaveName`** (`TopBar.tsx:349`): Silent failure on profile update — user gets no feedback.

**Worth considering:**

4. `get_profile` / `update_profile` do O(n) linear scans over the user store (keyed by email, searched by id).
5. Any team member can remove any other member including the creator, and removing the last member auto-deletes the team. Confirm this is the intended access model.

Overall the auth/teams implementation is solid — good JWT handling, proper bcrypt offloading to `spawn_blocking`, TOCTOU prevention on signup, and clean multi-tenant directory isolation in `user_context.rs`. The hooks violation is the only hard blocker.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #121 (previous HEAD: `d35c2cb`, new HEAD: `76b70d8`)_

cthulu-studio/src/components/TopBar.tsx
saltyskip, line 310: **Bug (React hooks violation)**: `if (!userEmail) return null;` on line 310 is an early return *before* the `useCallback` (line 312) and `useEffect` (lines 320, 325) calls below it. This violates the [Rules of Hooks](https://react.dev/reference/rules/rules-of-hooks) — hooks must be called in the same order on every render.

If `userEmail` transitions from non-null to null (e.g. on logout), React will crash with "Rendered fewer hooks than expected."

**Fix**: Move all hooks above the early return:
```tsx
function ProfileMenu() {
  const { logout, userEmail } = useAuth();
  const [open, setOpen] = useState(false);
  // ... all useState/useRef ...

  const loadData = useCallback(async () => { ... }, []);

  useEffect(() => { if (open) loadData(); }, [open, loadData]);
  useEffect(() => { /* mousedown handler */ }, [open]);

  if (!userEmail) return null; // <-- move early return here, after all hooks
  // ... rest of component
}
```

cthulu-backend/api/teams.rs
saltyskip, line 51: **Bug: Non-atomic write** — `TeamStore::save` uses direct `std::fs::write(path, json)`, which can leave a corrupted file on crash. The project's rule #9 (and `UserStore::save` in this same PR) requires atomic persistence via temp file + rename.

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
saltyskip, line 190: **Design concern**: Any team member can remove any other member (including the team creator), and a member can remove themselves with no confirmation. A user could accidentally (or maliciously) remove the creator, leaving the team with no "owner."

Consider at minimum preventing self-removal (force them to use "leave team" semantics or delete the team) and/or restricting removal to the `created_by` user.

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+    if (debounceRef.current) clearTimeout(debounceRef.current);
+    if (value.trim().length < 2) { setResults([]); return; }
+
+    setSearching(true);
+    debounceRef.current = setTimeout(async () => {
+      try {
+        const { users } = await api.searchUsers(value.trim());
+        setResults(users);
+      } catch { setResults([]); }
+      finally { setSearching(false); }
```
saltyskip, line 451: **Bug: `setSearching` race condition** — `setSearching(true)` is set synchronously, but the `false` happens inside the debounced `setTimeout` callback. If the user types fast:
1. Type 'a' → `setSearching(true)`, schedule timeout A
2. Type 'b' → clear timeout A, `setSearching(true)`, schedule timeout B
3. Timeout B fires → `setSearching(false)` ✓

This works in the happy path, but if the API call takes a while, the `finally` block sets `searching = false` even though the result may be stale (a newer timeout might be pending). Consider tracking the latest request with an abort controller or a request counter to avoid showing stale results.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,568 @@
...
+    axum::extract::State(state): axum::extract::State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.user_store.read().await;
+    let user = store.users.values().find(|u| u.id == auth.user_id);
+    match user {
+        Some(u) => (StatusCode::OK, Json(json!({
+            "id": u.id, "email": u.email, "name": u.name,
+            "avatar_url": u.avatar_url, "created_at": u.created_at,
+        }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
```
saltyskip, line 407: **Performance**: `get_profile` does `store.users.values().find(|u| u.id == auth.user_id)` — a linear scan because the HashMap is keyed by email, not user ID. Same pattern in `update_profile`. For a small user count this is fine, but consider adding a secondary `id -> email` index if this grows.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,568 @@
...
+    let Some(user) = store.users.get_mut(&email) else {
+        return (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" })));
+    };
+    if let Some(name) = body.name {
+        user.name = Some(name);
+    }
+    if let Some(avatar_url) = body.avatar_url {
+        user.avatar_url = Some(avatar_url);
+    }
+    let updated = user.clone();
```
saltyskip, line 439: Minor: No validation on `name` or `avatar_url` length. A client could store an arbitrarily large name string. Consider clamping to a reasonable max (e.g. 100 chars for name, 2048 for avatar URL).

saltyskip (changes requested):
## 🤖 Cthulu Review

**Must fix before merge:**

1. **React hooks violation in `ProfileMenu`** — `useCallback` and `useEffect` are called after an early `return null`, violating Rules of Hooks. This will crash on logout when `userEmail` transitions to null. Move all hooks above the conditional return.

2. **`TeamStore::save()` is not atomic** — Uses direct `std::fs::write` instead of temp-file + rename like `UserStore::save` does. This violates project rule #9 and risks data corruption on crash.

**Should fix:**

3. **Any member can remove any other member** (including the creator) from a team with no restrictions. Consider restricting removal to the `created_by` user or adding role-based checks.

**Minor / nice-to-have:**

4. User search debounce has a subtle race condition with stale results — consider abort controllers.
5. `get_profile` / `update_profile` do linear scans of the user store (keyed by email, searched by ID).
6. No length validation on profile `name` or `avatar_url` fields.

Overall the auth implementation is solid — good JWT handling, proper bcrypt offloading to `spawn_blocking`, atomic user store persistence, TOCTOU prevention on signup, and clean dev-mode bypass. The two blocking issues (hooks violation and non-atomic team save) are straightforward fixes.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #121 (previous HEAD: `76b70d8`, new HEAD: `042367f`)_

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+function ProfileMenu() {
+  const { logout, userEmail } = useAuth();
+  const [open, setOpen] = useState(false);
+  const [profile, setProfile] = useState<{ name: string | null; email: string } | null>(null);
+  const [teams, setTeams] = useState<{ id: string; name: string }[]>([]);
+  const [editingName, setEditingName] = useState(false);
+  const [nameInput, setNameInput] = useState("");
+  const [newTeamName, setNewTeamName] = useState("");
+  const [expandedTeam, setExpandedTeam] = useState<string | null>(null);
+  const [teamMembers, setTeamMembers] = useState<Record<string, { id: string; email: string; name: string | null }[]>>({});
```
saltyskip, line 301: **Bug: React hooks violation.** `useCallback` (line 303), `useEffect` (lines 311, 316) are called *after* the conditional `return null` on line 301. React requires all hooks to be called unconditionally at the top of the component — calling them after an early return means they won't be called on renders where `userEmail` is falsy, violating the Rules of Hooks and causing runtime crashes.

Fix: move the early return below all hook calls, or restructure to always call hooks first:

```tsx
function ProfileMenu() {
  const { logout, userEmail } = useAuth();
  const [open, setOpen] = useState(false);
  // ... all useState/useRef here ...

  const loadData = useCallback(async () => { ... }, []);

  useEffect(() => { if (open) loadData(); }, [open, loadData]);
  useEffect(() => { /* mousedown handler */ }, [open]);

  if (!userEmail) return null;  // early return AFTER all hooks
  // ... rest of component
}
```

cthulu-backend/api/teams.rs
saltyskip, line 51: **Bug: Non-atomic save.** `TeamStore::save` uses `std::fs::write` directly, which can corrupt the file if the process crashes mid-write. `UserStore::save` in `local_auth.rs` correctly uses temp-file + rename (lines 73-79). This should follow the same pattern per the project's rule #9 (atomic YAML persistence).

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
saltyskip, line 190: **Design issue: Any member can remove any other member**, including the team creator. There's no ownership or admin check. A team member could remove everyone else and become the sole member. Consider restricting removal to `created_by` or at least preventing removal of the creator:

```rust
if user_id == team.created_by && auth.user_id != team.created_by {
    return (StatusCode::FORBIDDEN, Json(json!({ "error": "Cannot remove team owner" })));
}
```

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,566 @@
...
+    axum::extract::State(state): axum::extract::State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.user_store.read().await;
+    let user = store.users.values().find(|u| u.id == auth.user_id);
+    match user {
+        Some(u) => (StatusCode::OK, Json(json!({
+            "id": u.id, "email": u.email, "name": u.name,
+            "avatar_url": u.avatar_url, "created_at": u.created_at,
+        }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
```
saltyskip, line 407: **Performance: O(n) scan on every profile request.** `users` is a `HashMap<String, StoredUser>` keyed by email, but `get_profile` and `update_profile` look up by `user_id` via `.values().find()`. With many users this becomes slow. Consider adding a secondary index (`HashMap<String, String>` mapping `id -> email`) or switching the primary key to `id`.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,566 @@
...
+    }
+
+    let store = state.user_store.read().await;
+    let results: Vec<Value> = store.users.values()
+        .filter(|u| {
+            u.email.to_lowercase().contains(&query)
+                || u.name.as_ref().map(|n| n.to_lowercase().contains(&query)).unwrap_or(false)
+        })
+        .take(10)
+        .map(|u| json!({ "id": u.id, "email": u.email, "name": u.name }))
```
saltyskip, line 465: **Missing pagination/rate limiting on user search.** The `.take(10)` limits results, but there's no rate limiting. An authenticated user could enumerate all users by searching single characters. Consider adding rate limiting at the middleware level, or at minimum requiring a minimum query length > 1.

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+    setEditingName(false);
+  };
+
+  const loadTeamMembers = async (teamId: string) => {
+    const { team } = await api.getTeam(teamId);
+    setTeamMembers((prev) => ({ ...prev, [teamId]: team.members as { id: string; email: string; name: string | null }[] }));
+  };
+
+  const handleExpandTeam = async (teamId: string) => {
+    if (expandedTeam === teamId) { setExpandedTeam(null); return; }
```
saltyskip, line 341: **Race condition in debounced search.** `setSearching(true)` fires immediately, but the actual fetch happens 300ms later. If the user clears the input (hits the `length < 2` branch), `setSearching` is never set back to `false` because the timeout is cleared. The "Searching..." indicator would remain stuck.

Also, concurrent timeouts can interleave — an older callback's `setSearching(false)` can fire after a newer search starts. Consider using an AbortController or a ref to track the latest request.

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+                      <div key={m.id} className="profile-member">
+                        <span className="profile-member-info">
+                          <span className="profile-member-name">{m.name || m.email.split("@")[0]}</span>
+                          <span className="profile-member-email">{m.email}</span>
+                        </span>
+                        <button className="profile-member-remove" onClick={() => handleRemoveMember(t.id, m.id)} title="Remove member">x</button>
+                      </div>
+                    ))}
+                    <UserSearchInput onSelect={async (_userId, email) => {
+                      try {
```
saltyskip, line 409: Minor: `memberError` state is shared across all teams but only displayed inside the expanded team. If you expand team A, get an error, collapse it, expand team B — the error from team A is still shown. Clear it when expanding a different team (which `handleExpandTeam` does via `setMemberError("")`, so this is partially handled, but the error persists if you stay in the same team and it was set from a previous failed add).

saltyskip (changes requested):
## 🤖 Cthulu Review

**Request changes** — one blocking bug, several issues to address.

### Blocking

- **React hooks violation in `ProfileMenu`**: `useCallback` and `useEffect` are called after an early `return null`, violating the Rules of Hooks. This will crash at runtime when `userEmail` is falsy. Move all hooks above the conditional return.

### Should Fix

- **Non-atomic `TeamStore::save`**: Uses `std::fs::write` directly — a crash mid-write corrupts `teams.json`. `UserStore::save` already does temp+rename correctly; `TeamStore` should match (project rule #9).
- **Any member can remove any member** including the team creator. No ownership check in `remove_member`.
- **Debounce race condition in `UserSearchInput`**: `setSearching(true)` fires immediately but the fetch is delayed 300ms. Clearing input leaves the "Searching..." indicator stuck.

### Minor / Consider

- `get_profile` and `update_profile` do O(n) linear scans by `user_id` on an email-keyed HashMap. Fine for now, will be a problem at scale.
- User search endpoint has no rate limiting — single-character queries can enumerate all users.
- Auth/teams implementation is solid overall: bcrypt with `spawn_blocking`, constant-time login, atomic user save, JWT secret with `0600` permissions, TOCTOU prevention on signup.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #121 (previous HEAD: `042367f`, new HEAD: `4111e98`)_

cthulu-studio/src/components/TopBar.tsx
saltyskip, line 325: **Bug (Rules of Hooks violation):** This early return is placed *before* `useCallback` (line 327) and `useEffect` (lines 336, 340). React requires hooks to be called in the same order on every render — an early return before hooks means they're called conditionally, which will crash React when `userEmail` transitions between null and non-null.

Move all hooks above this early return, or restructure so the conditional rendering doesn't short-circuit hook calls:

```tsx
// Move ALL hooks above the early return
const loadData = useCallback(async () => { ... }, []);
useEffect(() => { if (open) loadData(); }, [open, loadData]);
useEffect(() => { ... }, [open]);

if (!userEmail) return null;
```

cthulu-backend/api/teams.rs
saltyskip, line 51: **Bug (data loss risk):** `TeamStore::save` writes directly with `std::fs::write` — a crash mid-write will corrupt `teams.json`. `UserStore::save` in `local_auth.rs` correctly uses temp-file + rename (atomic write). This store should do the same:

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

This also aligns with the project rule #9 (atomic YAML persistence pattern).

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,569 @@
...
+    axum::extract::State(state): axum::extract::State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.user_store.read().await;
+    let user = store.users.values().find(|u| u.id == auth.user_id);
+    match user {
+        Some(u) => (StatusCode::OK, Json(json!({
+            "id": u.id, "email": u.email, "name": u.name,
+            "avatar_url": u.avatar_url, "created_at": u.created_at,
+        }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
```
saltyskip, line 407: **Performance:** `get_profile` does a linear scan (`values().find()`) over all users to find by ID, but `UserStore` is keyed by email. Same issue in `update_profile` (line 429). Consider adding a secondary index (`id -> email` HashMap) or changing the primary key to user ID, since profile lookups by ID happen on every authenticated page load.

cthulu-studio/src/components/TopBar.tsx
saltyskip, line 466: **Bug (race condition):** `setSearching(true)` is set synchronously, but the actual search fires 300ms later via `setTimeout`. If the user types then clears the input (triggering the `value.trim().length < 2` early return on line 463), `searching` stays `true` forever because the timeout was cleared before it could call `setSearching(false)`.

Fix: also `setSearching(false)` in the early return branch:
```tsx
if (value.trim().length < 2) { setResults([]); setSearching(false); return; }
```

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,569 @@
...
+
+    // Constant-time: always run bcrypt verify to prevent user enumeration via timing.
+    // If user doesn't exist, verify against a dummy hash (same cost as real verify).
+    let dummy_hash = "$2b$12$000000000000000000000u2a5OJr0FkDxcMkGCuaLxMPqOIZJcMK";
+    let password = body.password.clone();
+    let hash = user.as_ref().map(|u| u.password_hash.clone())
+        .unwrap_or_else(|| dummy_hash.to_string());
+    let valid = tokio::task::spawn_blocking(move || bcrypt::verify(&password, &hash).unwrap_or(false))
+        .await
+        .unwrap_or(false);
```
saltyskip, line 365: **Nit:** The dummy hash is a good idea for timing-safe enumeration prevention, but this hardcoded value must be a valid bcrypt hash that `bcrypt::verify` accepts without erroring. If it's not a well-formed hash, `verify` returns `Err` (caught by `unwrap_or(false)` so no crash, but the timing may differ from a real verify). Consider generating this once at startup with `bcrypt::hash("dummy", 12)` to guarantee valid format and consistent timing.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
```
saltyskip, line 189: **Design concern:** Any team member can remove any other member (including the team creator). When the last member is removed (line 191-193), the team is silently deleted. A member could also remove *themselves*, destroying the team if they're the last one. Consider:
1. Preventing removal of the team creator, or
2. At minimum, confirming this is intentional in the API docs since it's a "no roles" model.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,569 @@
...
+    if let Some(query) = parts.uri.query() {
+        for pair in query.split('&') {
+            if let Some(token) = pair.strip_prefix("token=") {
+                return Some(percent_decode_str(token).decode_utf8_lossy().into_owned());
+            }
+        }
+    }
+
+    None
+}
```
saltyskip, line 243: **Security note:** The `?token=` query param fallback for SSE/WebSocket is reasonable, but tokens in URLs get logged in server access logs, browser history, and proxy logs. Consider documenting this risk or adding a log warning when token auth is used via query param.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Must fix before merge (2 bugs):**

1. **Rules of Hooks violation in `ProfileMenu`** — Early return (`if (!userEmail) return null`) on line 325 comes *before* `useCallback`/`useEffect` hooks. This will crash React at runtime when the auth state changes. Move all hooks above the early return.

2. **`TeamStore::save` is not atomic** — Uses `std::fs::write` directly, unlike `UserStore::save` which correctly uses temp-file + rename. A crash mid-write will corrupt `teams.json`. This also violates project rule #9 (atomic persistence).

**Should fix:**

3. **`UserSearchInput` race condition** — `searching` state gets stuck as `true` when user clears input after typing, because the early return doesn't reset it.

4. **`get_profile` / `update_profile` linear scan** — O(n) scan over all users on every authenticated page load because the store is keyed by email but lookups are by user ID.

**Minor / design notes:**

5. Dummy bcrypt hash for timing-safe login should be validated or generated at startup to ensure consistent timing.
6. Any team member can remove any other member (including creator) and deleting the last member silently destroys the team — confirm this is intentional for the "flat" model.
7. Query-param token fallback (`?token=`) for SSE is functional but tokens in URLs leak to logs/history.

Overall the auth and teams implementation is solid — good use of bcrypt with `spawn_blocking`, CSPRNG for JWT secrets, proper TOCTOU prevention on signup, and timing-safe login. The two blocking issues are straightforward fixes.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #121 (previous HEAD: `4111e98`, new HEAD: `11165a4`)_

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+  const [newTeamName, setNewTeamName] = useState("");
+  const [expandedTeam, setExpandedTeam] = useState<string | null>(null);
+  const [teamMembers, setTeamMembers] = useState<Record<string, { id: string; email: string; name: string | null }[]>>({});
+  const [memberError, setMemberError] = useState("");
+  const menuRef = useRef<HTMLDivElement>(null);
+
+  if (!userEmail) return null;
+
+  const loadData = useCallback(async () => {
+    try {
```
saltyskip, line 308: **Bug: React Rules of Hooks violation.** The early `return null` on line 308 is placed _after_ some hooks (`useState`, `useRef`) but _before_ others (`useCallback` on line 310, `useEffect` on lines 319/323). React requires hooks to be called in the same order on every render — conditional returns between hooks violate this.

While this may not crash today because `VITE_AUTH_ENABLED` is static, the linter will flag it and it's a latent bug if anything ever changes `userEmail` dynamically (e.g. logout/login cycle within the same mount).

**Fix:** Move all hooks above the early return:
```tsx
function ProfileMenu() {
  const { logout, userEmail } = useAuth();
  const [open, setOpen] = useState(false);
  // ... all other useState, useRef, useCallback, useEffect hooks ...
  
  if (!userEmail) return null;
  
  // ... render logic only below here
}
```

cthulu-backend/api/teams.rs
saltyskip, line 51: **Non-atomic save — data corruption risk on crash.** `UserStore::save` correctly uses temp-file + rename (crash-safe atomic write), but `TeamStore::save` uses bare `std::fs::write`. If the process crashes mid-write, `teams.json` will be truncated/corrupt.

This also violates the project rule #9 ("Atomic YAML persistence: Always write sessions via temp file + rename"). The same pattern should apply to all persisted JSON.

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/local_auth.rs
saltyskip, line 407: Minor: `get_profile` and `update_profile` both do `store.users.values().find(|u| u.id == auth.user_id)` — an O(n) linear scan because the HashMap is keyed by email, not user ID. Consider adding a secondary `id_to_email: HashMap<String, String>` index if user counts grow. Not blocking for a self-hosted tool, but worth noting.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    let store = state.team_store.read().await;
+    match store.teams.get(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            let user_store = state.user_store.read().await;
+            let members: Vec<Value> = team.members.iter().map(|uid| {
+                let user = user_store.users.values().find(|u| u.id == *uid);
+                json!({
+                    "id": uid,
+                    "email": user.map(|u| u.email.as_str()).unwrap_or("unknown"),
+                    "name": user.and_then(|u| u.name.as_deref()),
```
saltyskip, line 121: Nit: `get_team` holds `team_store.read()` while also acquiring `user_store.read()` (line 121). `add_member` acquires them in the opposite order (user_store first, team_store second). While both are read locks here and won't deadlock with each other, the inconsistent ordering could become a problem if either is refactored to use a write lock. Consider standardizing the acquisition order across all handlers (e.g., always user_store then team_store).

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+  const handleChange = (value: string) => {
+    setQuery(value);
+    if (debounceRef.current) clearTimeout(debounceRef.current);
+    if (value.trim().length < 2) { setResults([]); return; }
+
+    setSearching(true);
+    debounceRef.current = setTimeout(async () => {
+      try {
+        const { users } = await api.searchUsers(value.trim());
+        setResults(users);
```
saltyskip, line 449: Minor: the debounce sets `setSearching(true)` immediately on keystroke (line 448) but the actual fetch happens after the 300ms timeout. If the user types and then stops before 2 chars, `searching` stays `true` forever because the early return on line 446 skips the timeout that would set it `false`. Add `setSearching(false)` to the early return path:
```tsx
if (value.trim().length < 2) { setResults([]); setSearching(false); return; }
```

saltyskip (changes requested):
## 🤖 Cthulu Review

Two issues to fix before merging:

1. **React hooks violation in `ProfileMenu`** — Early `return null` on line 308 is placed between hooks (`useState`/`useRef` above, `useCallback`/`useEffect` below). This violates React's Rules of Hooks. Move all hooks above the conditional return.

2. **`TeamStore::save` is not atomic** — Uses bare `std::fs::write` while `UserStore::save` correctly uses temp-file + rename. A crash mid-write would corrupt `teams.json`. Apply the same atomic pattern.

### Minor notes (non-blocking)
- `get_profile`/`update_profile` do O(n) scans since the user HashMap is keyed by email, not ID
- Lock acquisition order for `team_store`/`user_store` is inconsistent between `get_team` and `add_member` — fine with read locks today, risky if refactored
- `UserSearchInput` debounce leaves `searching` stuck as `true` if query is cleared below 2 chars

Auth implementation overall looks solid — constant-time login comparison, CSPRNG JWT secret with 0600 perms, atomic user store writes, and clean token extraction with query-param fallback for SSE.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #121 (previous HEAD: `11165a4`, new HEAD: `ed4b570`)_

cthulu-studio/src/components/TopBar.tsx
saltyskip, line 301: **Bug: Rules of Hooks violation.** `if (!userEmail) return null;` appears before `useCallback` and `useEffect` hooks (lines 310-326). React requires hooks to be called in the same order every render — conditional early returns before hooks will crash in dev mode and cause undefined behavior in production.

Fix: move the early return **after** all hooks:
```tsx
function ProfileMenu() {
  const { logout, userEmail } = useAuth();
  const [open, setOpen] = useState(false);
  // ... all other hooks ...

  if (!userEmail) return null;

  // ... render logic ...
}
```

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+  const debounceRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  const handleChange = (value: string) => {
+    setQuery(value);
+    if (debounceRef.current) clearTimeout(debounceRef.current);
+    if (value.trim().length < 2) { setResults([]); return; }
+
+    setSearching(true);
+    debounceRef.current = setTimeout(async () => {
+      try {
```
saltyskip, line 447: **Bug: `searching` state can get stuck as `true`.** `setSearching(true)` is called immediately (line 446), but `setSearching(false)` only runs inside the `setTimeout` callback. If the user types again within 300ms, the old timeout is cleared — the old `setSearching(false)` never fires, and the new `setSearching(true)` overwrites it. This means "Searching..." can appear permanently.

Fix: set `setSearching(true)` inside the timeout callback, right before the fetch:
```tsx
setSearching(true); // remove from here
debounceRef.current = setTimeout(async () => {
  setSearching(true); // add here instead
  try { ... }
  finally { setSearching(false); }
}, 300);
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
```
saltyskip, line 48: **Convention violation: non-atomic file write.** CLAUDE.md rule 9 requires atomic persistence via temp file + rename. `std::fs::write(path, json)` can corrupt the file if the process crashes mid-write.

Should be:
```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(tmp, path)
}
```

cthulu-studio/src/api/client.ts
```diff
@@ -47,18 +80,33 @@ async function apiFetch<T>(
...
         ...options.headers,
       },
     });
 
     const elapsed = Math.round(performance.now() - start);
 
     if (!res.ok) {
+      // Auto-logout on 401 (expired/invalid token)
+      if (res.status === 401 && getTokenFn) {
+        localStorage.removeItem("cthulu_auth_token");
```
saltyskip, line 107: **UX concern: `window.location.reload()` on 401 can lose unsaved work.** If a user has unsaved flow edits and the token expires, this silently discards everything. Consider dispatching a logout event or showing a re-auth modal instead of a hard reload.

cthulu-studio/src/api/client.ts
```diff
@@ -37,6 +37,39 @@ export function getServerUrl(): string {
...
+
+/** Append ?token= to a URL for EventSource connections (which can't set headers). */
+export function withAuthToken(url: string): string {
+  const token = getAuthTokenSync();
+  if (!token) return url;
+  const sep = url.includes("?") ? "&" : "?";
+  return `${url}${sep}token=${encodeURIComponent(token)}`;
+}
+
+/** Get auth headers for fetch-based streaming (non-apiFetch callers). */
```
saltyskip, line 67: **Security note: auth token in URL query params.** `withAuthToken()` appends tokens to URLs for EventSource connections. Tokens in URLs leak via server logs, browser history, and Referer headers. This is a known EventSource limitation, but make sure:
1. Server-side SSE endpoints actually parse `?token=` (not shown in this PR)
2. Tokens are short-lived
3. Server access logs are sanitized

cthulu-backend/api/local_auth.rs
```diff
@@ -389,6 +391,83 @@ async fn login(
...
+    name: Option<String>,
+    avatar_url: Option<String>,
+}
+
+async fn update_profile(
+    auth: AuthUser,
+    axum::extract::State(state): axum::extract::State<AppState>,
+    Json(body): Json<UpdateProfileRequest>,
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.user_store.write().await;
```
saltyskip, line 422: **Minor inefficiency:** `get_profile` does `store.users.values().find(|u| u.id == auth.user_id)` — O(n) scan since the store is keyed by email. Fine for small user counts, but if `AuthUser` already has the email, you could do a direct HashMap lookup.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        self.teams.values()
```
saltyskip, line 49: **Silent error swallowing.** `let _ = store.save(...)` silently ignores disk write failures in all handlers (`create_team`, `add_member`, `remove_member`). The in-memory state diverges from disk with no indication to the user. At minimum, log the error. Better: return a 500 if save fails.

saltyskip (changes requested):
## 🤖 Cthulu Review

**2 bugs, 1 convention violation, and a few concerns to address before merging.**

### Must fix

1. **Rules of Hooks violation in `ProfileMenu`** — Early return `if (!userEmail) return null` on line 301 comes before `useCallback`/`useEffect` hooks. This will crash in React dev mode. Move the guard after all hooks.

2. **`searching` state stuck in `UserSearchInput`** — `setSearching(true)` fires immediately but `setSearching(false)` is inside the debounced timeout. Clearing the timeout leaves `searching` permanently true, showing "Searching..." forever.

3. **Non-atomic `TeamStore::save`** — Uses `std::fs::write` directly, violating CLAUDE.md rule 9 (atomic persistence via temp file + rename). Should write to `.json.tmp` then rename.

### Should fix

4. **`window.location.reload()` on 401** — Hard reload on expired token discards unsaved work with no warning. Consider a re-auth modal or event-based logout.

5. **Silent `let _ = store.save()`** — Disk write failures are ignored in all team/profile handlers. In-memory state diverges from disk silently. At minimum log the error.

6. **Token-in-URL for EventSource** — Confirm server-side SSE endpoints actually parse `?token=` query param (not visible in this PR's diff).

### Nits

7. `get_profile` does O(n) scan by user ID when store is keyed by email.

---

_Automated review by Cthulu_

PR #122, opened 2026-03-20
feat: team sessions — shared agents with multi-user broadcast

MPRR-Pandu:
## Summary

Depends on: #121 (profile + teams UI) — merge that first.

- Add `team_id` field to Agent struct — team-owned agents
- Team membership check on chat endpoint (403 for non-members)
- Broadcast `user_message` SSE event with sender attribution
- Frontend handles `user_message` event in chat stream
- Sidebar splits agents into "Agents" (personal) + per-team sections
- Agent create picker: "+" dropdown offers Personal Agent or Team Agent
- Visual distinction: accent border + "team" badge on team agents

## How It Works

1. Create a team agent via "+" picker (select a team)
2. All team members see it in their team's sidebar section
3. Any member sends a message — `user_message` event broadcasts
4. Other members join via reconnect endpoint (spectate in real-time)
5. One sender at a time (existing 409 while busy)

## Files changed (7 files, ~146 lines)
- `agents/mod.rs` — `team_id` field on Agent
- `chat.rs` — AuthUser extractor, team check, user_message broadcast
- `Sidebar.tsx` — team sections + agent create picker
- `useAgentChat.ts` — handle `user_message` SSE event
- `flow.ts` — `team_id` on Agent/AgentSummary types
- `client.ts` — `team_id` on createAgent
- `styles.css` — team agent visual styles

## Verification
- `cargo check` — 0 errors
- `cargo test` — 319 passed

## Dependency chain
**PR F (6/6)** — final PR, merge after #121

```
main → #117 → #118 → #119 + #120 → #121 → THIS PR F (team sessions)
```

### DIFF
--- a/cthulu-backend/agents/mod.rs
+++ b/cthulu-backend/agents/mod.rs
@@ -96,6 +96,9 @@ pub struct Agent {
     /// If true, this agent is only usable as a sub-agent and is hidden from the UI agent list.
     #[serde(default)]
     pub subagent_only: bool,
+    /// When set, this agent belongs to a team. All team members can access it.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub team_id: Option<String>,
     pub created_at: DateTime<Utc>,
     pub updated_at: DateTime<Utc>,
 }
@@ -214,6 +217,7 @@ impl AgentBuilder<Ready> {
             hooks: self.hooks,
             subagents: self.subagents,
             subagent_only: self.subagent_only,
+            team_id: None,
             created_at: now,
             updated_at: now,
         }
--- a/cthulu-backend/api/agents/chat.rs
+++ b/cthulu-backend/api/agents/chat.rs
@@ -781,6 +781,7 @@ fn chat_sdk_stream(
 
 /// POST /agents/{id}/chat — SSE stream for agent chat
 pub(crate) async fn chat(
+    auth: crate::api::local_auth::AuthUser,
     State(state): State<AppState>,
     Path(id): Path<String>,
     Json(body): Json<ChatRequest>,
@@ -792,6 +793,15 @@ pub(crate) async fn chat(
         (StatusCode::NOT_FOUND, Json(json!({ "error": "agent not found" })))
     })?;
 
+    // If team agent, verify sender is a team member
+    if let Some(ref team_id) = agent.team_id {
+        let ts = state.team_store.read().await;
+        if !ts.teams.get(team_id).map(|t| t.members.contains(&auth.user_id)).unwrap_or(false) {
+            return Err((StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))));
+        }
+    }
+
+    let sender_user_id = auth.user_id.clone();
     let prompt = body.prompt;
     let images = body.images.unwrap_or_default();
     if prompt.trim().is_empty() && images.is_empty() {
@@ -1208,6 +1218,8 @@ pub(crate) async fn chat(
     let live_processes = state.live_processes.clone();
     let session_streams = state.session_streams.clone();
     let chat_event_buffers = state.chat_event_buffers.clone();
+    let sender_id_for_bc = sender_user_id.clone();
+    let prompt_for_bc = prompt.chars().take(200).collect::<String>();
 
     let stream = async_stream::stream! {
         use std::process::Stdio;
@@ -1438,6 +1450,16 @@ pub(crate) async fn chat(
                 proc_key = %proc_key_for_stream,
                 "[RECONNECT-DEBUG] Created event buffer"
             );
+        }
+
+        // Broadcast sender attribution for team session spectators
+        {
+            let evt = format!("user_message:{}", serde_json::to_string(&json!({
+                "user_id": sender_id_for_bc, "text": prompt_for_bc,
+            })).unwrap_or_default());
+            let _ = bc_tx.send(evt.clone());
+            let mut bufs = chat_event_buffers.lock().await;
+            if let Some(b) = bufs.get_mut(&proc_key_for_stream) { b.push(evt); }
         }
 
         // Spawn the background reader task
--- a/cthulu-backend/api/local_auth.rs
+++ b/cthulu-backend/api/local_auth.rs
@@ -7,7 +7,7 @@
 //! and all requests get a hardcoded "dev_user" identity.
 
 use axum::extract::FromRequestParts;
-use axum::routing::post;
+use axum::routing::{get, post, put};
 use axum::{Json, Router};
 use hyper::header::AUTHORIZATION;
 use hyper::http::request::Parts;
@@ -260,6 +260,8 @@ pub fn router() -> Router<AppState> {
     Router::new()
         .route("/auth/signup", post(signup))
         .route("/auth/login", post(login))
+        .route("/auth/me", get(get_profile).put(update_profile))
+        .route("/auth/users/search", get(search_users))
 }
 
 async fn signup(
@@ -387,6 +389,83 @@ async fn login(
             "user": { "id": user.id, "email": user.email }
         })),
     )
+}
+
+// ── Profile Handlers ─────────────────────────────────────────
+
+async fn get_profile(
+    auth: AuthUser,
+    axum::extract::State(state): axum::extract::State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.user_store.read().await;
+    let user = store.users.values().find(|u| u.id == auth.user_id);
+    match user {
+        Some(u) => (StatusCode::OK, Json(json!({
+            "id": u.id, "email": u.email, "name": u.name,
+            "avatar_url": u.avatar_url, "created_at": u.created_at,
+        }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
+    }
+}
+
+#[derive(Deserialize)]
+struct UpdateProfileRequest {
+    name: Option<String>,
+    avatar_url: Option<String>,
+}
+
+async fn update_profile(
+    auth: AuthUser,
+    axum::extract::State(state): axum::extract::State<AppState>,
+    Json(body): Json<UpdateProfileRequest>,
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.user_store.write().await;
+    let email_key = store.users.iter()
+        .find(|(_, u)| u.id == auth.user_id)
+        .map(|(email, _)| email.clone());
+
+    let Some(email) = email_key else {
+        return (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" })));
+    };
+    let Some(user) = store.users.get_mut(&email) else {
+        return (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" })));
+    };
+    if let Some(name) = body.name {
+        user.name = Some(name);
+    }
+    if let Some(avatar_url) = body.avatar_url {
+        user.avatar_url = Some(avatar_url);
+    }
+    let updated = user.clone();
+    let _ = store.save(&state.data_dir);
+
+    (StatusCode::OK, Json(json!({
+        "id": updated.id, "email": updated.email,
+        "name": updated.name, "avatar_url": updated.avatar_url,
+    })))
+}
+
+async fn search_users(
+    _auth: AuthUser,
+    axum::extract::State(state): axum::extract::State<AppState>,
+    axum::extract::Query(params): axum::extract::Query<std::collections::HashMap<String, String>>,
+) -> (StatusCode, Json<Value>) {
+    let query = params.get("q").map(|s| s.trim().to_lowercase()).unwrap_or_default();
+    if query.is_empty() {
+        return (StatusCode::OK, Json(json!({ "users": [] })));
+    }
+
+    let store = state.user_store.read().await;
+    let results: Vec<Value> = store.users.values()
+        .filter(|u| {
+            u.email.to_lowercase().contains(&query)
+                || u.name.as_ref().map(|n| n.to_lowercase().contains(&query)).unwrap_or(false)
+        })
+        .take(10)
+        .map(|u| json!({ "id": u.id, "email": u.email, "name": u.name }))
+        .collect();
+
+    (StatusCode::OK, Json(json!({ "users": results })))
 }
 
 // ── Tests ────────────────────────────────────────────────────
--- a/cthulu-backend/api/teams.rs
+++ b/cthulu-backend/api/teams.rs
@@ -1,0 +1,240 @@
+//! Teams: flat groups of users. No roles — all members are equal.
+//! Stored in `{data_dir}/teams.json`.
+
+use axum::extract::{Path, State};
+use axum::routing::{delete, get, post};
+use axum::{Json, Router};
+use hyper::StatusCode;
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Value};
+use std::collections::HashMap;
+use std::path::PathBuf;
+
+use crate::api::local_auth::AuthUser;
+use crate::api::AppState;
+
+// ── Team Model ───────────────────────────────────────────────
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Team {
+    pub id: String,
+    pub name: String,
+    pub created_by: String,
+    pub members: Vec<String>,
+    pub created_at: String,
+}
+
+/// In-memory team store backed by `{data_dir}/teams.json`.
+pub struct TeamStore {
+    pub teams: HashMap<String, Team>,
+}
+
+impl TeamStore {
+    pub fn load(data_dir: &PathBuf) -> Self {
+        let path = data_dir.join("teams.json");
+        let teams = match std::fs::read_to_string(&path) {
+            Ok(contents) => serde_json::from_str(&contents).unwrap_or_default(),
+            Err(_) => HashMap::new(),
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
+        self.teams.values()
+            .filter(|t| t.members.contains(&user_id.to_string()))
+            .collect()
+    }
+}
+
+// ── Routes ───────────────────────────────────────────────────
+
+pub fn router() -> Router<AppState> {
+    Router::new()
+        .route("/teams", get(list_teams).post(create_team))
+        .route("/teams/{id}", get(get_team))
+        .route("/teams/{id}/members", post(add_member))
+        .route("/teams/{id}/members/{user_id}", delete(remove_member))
+}
+
+// ── Handlers ─────────────────────────────────────────────────
+
+async fn list_teams(
+    auth: AuthUser,
+    State(state): State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.team_store.read().await;
+    let teams: Vec<&Team> = store.teams_for_user(&auth.user_id);
+    (StatusCode::OK, Json(json!({ "teams": teams })))
+}
+
+#[derive(Deserialize)]
+struct CreateTeamRequest {
+    name: String,
+}
+
+async fn create_team(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Json(body): Json<CreateTeamRequest>,
+) -> (StatusCode, Json<Value>) {
+    let name = body.name.trim().to_string();
+    if name.is_empty() {
+        return (StatusCode::BAD_REQUEST, Json(json!({ "error": "Team name required" })));
+    }
+
+    let team = Team {
+        id: uuid::Uuid::new_v4().to_string().replace('-', "_"),
+        name,
+        created_by: auth.user_id.clone(),
+        members: vec![auth.user_id.clone()],
+        created_at: chrono::Utc::now().to_rfc3339(),
+    };
+
+    let mut store = state.team_store.write().await;
+    let response = json!({ "team": &team });
+    store.teams.insert(team.id.clone(), team);
+    let _ = store.save(&state.data_dir);
+
+    (StatusCode::CREATED, Json(response))
+}
+
+async fn get_team(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path(id): Path<String>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.team_store.read().await;
+    match store.teams.get(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            let user_store = state.user_store.read().await;
+            let members: Vec<Value> = team.members.iter().map(|uid| {
+                let user = user_store.users.values().find(|u| u.id == *uid);
+                json!({
+                    "id": uid,
+                    "email": user.map(|u| u.email.as_str()).unwrap_or("unknown"),
+                    "name": user.and_then(|u| u.name.as_deref()),
+                })
+            }).collect();
+
+            (StatusCode::OK, Json(json!({
+                "team": {
+                    "id": team.id,
+                    "name": team.name,
+                    "created_by": team.created_by,
+                    "members": members,
+                    "created_at": team.created_at,
+                }
+            })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
+#[derive(Deserialize)]
+struct AddMemberRequest {
+    email: String,
+}
+
+async fn add_member(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path(id): Path<String>,
+    Json(body): Json<AddMemberRequest>,
+) -> (StatusCode, Json<Value>) {
+    let email = body.email.trim().to_lowercase();
+
+    let user_store = state.user_store.read().await;
+    let target_user = user_store.find_by_email(&email);
+    let target_id = match target_user {
+        Some(u) => u.id.clone(),
+        None => return (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
+    };
+    drop(user_store);
+
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            if team.members.contains(&target_id) {
+                return (StatusCode::CONFLICT, Json(json!({ "error": "Already a member" })));
+            }
+            team.members.push(target_id);
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
+async fn remove_member(
+    auth: AuthUser,
+    State(state): State<AppState>,
+    Path((id, user_id)): Path<(String, String)>,
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
+// ── Tests ────────────────────────────────────────────────────
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn team_store_roundtrip() {
+        let tmp = tempfile::TempDir::new().unwrap();
+        let dir = tmp.path().to_path_buf();
+
+        let mut store = TeamStore { teams: HashMap::new() };
+        store.teams.insert("t1".into(), Team {
+            id: "t1".into(),
+            name: "Alpha".into(),
+            created_by: "u1".into(),
+            members: vec!["u1".into(), "u2".into()],
+            created_at: "2026-01-01".into(),
+        });
+        store.save(&dir).unwrap();
+
+        let loaded = TeamStore::load(&dir);
+        assert_eq!(loaded.teams.len(), 1);
+        assert_eq!(loaded.teams["t1"].name, "Alpha");
+        assert_eq!(loaded.teams["t1"].members.len(), 2);
+    }
+
+    #[test]
+    fn teams_for_user_filters_correctly() {
+        let mut store = TeamStore { teams: HashMap::new() };
+        store.teams.insert("t1".into(), Team {
+            id: "t1".into(), name: "A".into(), created_by: "u1".into(),
+            members: vec!["u1".into(), "u2".into()], created_at: "".into(),
+        });
+        store.teams.insert("t2".into(), Team {
+            id: "t2".into(), name: "B".into(), created_by: "u3".into(),
+            members: vec!["u3".into()], created_at: "".into(),
+        });
+
+        assert_eq!(store.teams_for_user("u1").len(), 1);
+        assert_eq!(store.teams_for_user("u2").len(), 1);
+        assert_eq!(store.teams_for_user("u3").len(), 1);
+        assert_eq!(store.teams_for_user("u99").len(), 0);
+    }
+}
--- a/cthulu-studio/src/api/client.ts
+++ b/cthulu-studio/src/api/client.ts
@@ -37,6 +37,39 @@ export function getServerUrl(): string {
   return getBaseUrl();
 }
 
+// ── Auth token injection ─────────────────────────────────────
+
+let getTokenFn: (() => Promise<string | null>) | null = null;
+
+/** Called by AuthGate to wire token getter into all API requests. */
+export function setAuthTokenGetter(fn: (() => Promise<string | null>) | null) {
+  getTokenFn = fn;
+}
+
+/** Get the current auth token (if available). Used by SSE EventSource callers. */
+export async function getAuthToken(): Promise<string | null> {
+  return getTokenFn ? getTokenFn() : null;
+}
+
+/** Get auth token synchronously from localStorage (for EventSource URLs). */
+export function getAuthTokenSync(): string | null {
+  return localStorage.getItem("cthulu_auth_token");
+}
+
+/** Append ?token= to a URL for EventSource connections (which can't set headers). */
+export function withAuthToken(url: string): string {
+  const token = getAuthTokenSync();
+  if (!token) return url;
+  const sep = url.includes("?") ? "&" : "?";
+  return `${url}${sep}token=${encodeURIComponent(token)}`;
+}
+
+/** Get auth headers for fetch-based streaming (non-apiFetch callers). */
+export async function getAuthHeaders(): Promise<Record<string, string>> {
+  const token = await getAuthToken();
+  return token ? { Authorization: `Bearer ${token}` } : {};
+}
+
 async function apiFetch<T>(
   path: string,
   options: RequestInit = {}
@@ -47,11 +80,21 @@ async function apiFetch<T>(
   log("http", `${method} ${path}`);
   const start = performance.now();
 
+  // Attach auth token if available
+  const authHeaders: Record<string, string> = {};
+  if (getTokenFn) {
+    const token = await getTokenFn();
+    if (token) {
+      authHeaders["Authorization"] = `Bearer ${token}`;
+    }
+  }
+
   try {
     const res = await fetch(url, {
       ...options,
       headers: {
         "Content-Type": "application/json",
+        ...authHeaders,
         ...options.headers,
       },
     });
@@ -59,6 +102,11 @@ async function apiFetch<T>(
     const elapsed = Math.round(performance.now() - start);
 
     if (!res.ok) {
+      // Auto-logout on 401 (expired/invalid token)
+      if (res.status === 401 && getTokenFn) {
+        localStorage.removeItem("cthulu_auth_token");
+        window.location.reload();
+      }
       const body = await res.text();
       log("error", `${method} ${path} -> ${res.status} (${elapsed}ms)`, body);
       throw new Error(`API error ${res.status}: ${body}`);
@@ -326,7 +374,7 @@ export function streamSessionLog(
   onLine: (line: string) => void,
   onDone: () => void
 ): () => void {
-  const url = `${getBaseUrl()}/api/agents/${agentId}/sessions/${sessionId}/stream`;
+  const url = withAuthToken(`${getBaseUrl()}/api/agents/${agentId}/sessions/${sessionId}/stream`);
   const source = new EventSource(url);
 
   source.addEventListener("line", (e: MessageEvent) => {
@@ -540,6 +588,7 @@ export async function createAgent(data: {
   permissions?: string[];
   append_system_prompt?: string | null;
   working_dir?: string | null;
+  team_id?: string | null;
 }): Promise<{ id: string }> {
   return apiFetch<{ id: string }>("/agents", {
     method: "POST",
@@ -586,7 +635,7 @@ export interface ResourceChangeEvent {
 export function subscribeToChanges(
   onEvent: (event: ResourceChangeEvent) => void
 ): () => void {
-  const url = `${getBaseUrl()}/api/changes`;
+  const url = withAuthToken(`${getBaseUrl()}/api/changes`);
   const source = new EventSource(url);
 
   const handler = (e: MessageEvent) => {
@@ -706,4 +755,68 @@ export async function getDashboardSummary(
     method: "POST",
     body: JSON.stringify({ channels }),
   });
+}
+
+// ── Profile ──────────────────────────────────────────────────
+
+export interface UserProfile {
+  id: string;
+  email: string;
+  name: string | null;
+  avatar_url: string | null;
+  created_at: string;
+}
+
+export async function getProfile(): Promise<UserProfile> {
+  return apiFetch<UserProfile>("/auth/me");
+}
+
+export async function updateProfile(data: { name?: string; avatar_url?: string }): Promise<UserProfile> {
+  return apiFetch<UserProfile>("/auth/me", { method: "PUT", body: JSON.stringify(data) });
+}
+
+export interface UserSearchResult {
+  id: string;
+  email: string;
+  name: string | null;
+}
+
+export async function searchUsers(query: string): Promise<{ users: UserSearchResult[] }> {
+  return apiFetch<{ users: UserSearchResult[] }>(`/auth/users/search?q=${encodeURIComponent(query)}`);
+}
+
+// ── Teams ────────────────────────────────────────────────────
+
+export interface TeamMember {
+  id: string;
+  email: string;
+  name: string | null;
+}
+
+export interface Team {
+  id: string;
+  name: string;
+  created_by: string;
+  members: TeamMember[] | string[];
+  created_at: string;
+}
+
+export async function listTeams(): Promise<{ teams: Team[] }> {
+  return apiFetch<{ teams: Team[] }>("/teams");
+}
+
+export async function createTeam(name: string): Promise<{ team: Team }> {
+  return apiFetch<{ team: Team }>("/teams", { method: "POST", body: JSON.stringify({ name }) });
+}
+
+export async function getTeam(id: string): Promise<{ team: Team }> {
+  return apiFetch<{ team: Team }>(`/teams/${id}`);
+}
+
+export async function addTeamMember(teamId: string, email: string): Promise<{ ok: boolean }> {
+  return apiFetch<{ ok: boolean }>(`/teams/${teamId}/members`, { method: "POST", body: JSON.stringify({ email }) });
+}
+
+export async function removeTeamMember(teamId: string, userId: string): Promise<{ ok: boolean }> {
+  return apiFetch<{ ok: boolean }>(`/teams/${teamId}/members/${userId}`, { method: "DELETE" });
 }
--- a/cthulu-studio/src/components/AuthGate.tsx
+++ b/cthulu-studio/src/components/AuthGate.tsx
@@ -1,0 +1,176 @@
+import { useState, useCallback, createContext, useContext } from "react";
+import { setAuthTokenGetter, getServerUrl } from "../api/client";
+
+// ── Auth context so any component can call logout ────────────
+
+interface AuthContextValue {
+  logout: () => void;
+  userEmail: string | null;
+}
+
+const AuthContext = createContext<AuthContextValue>({
+  logout: () => {},
+  userEmail: null,
+});
+
+/** Hook for any component to access logout + user info. */
+export function useAuth() {
+  return useContext(AuthContext);
+}
+
+interface AuthGateProps {
+  children: React.ReactNode;
+}
+
+/** Decode base64url to string (handles missing padding). */
+function b64urlDecode(str: string): string {
+  const padded = str.replace(/-/g, "+").replace(/_/g, "/") + "=".repeat((4 - (str.length % 4)) % 4);
+  return atob(padded);
+}
+
+/** Parse JWT payload and return claims, or null if unparseable. */
+function parseToken(token: string): { email?: string; exp?: number } | null {
+  try {
+    const payload = token.split(".")[1];
+    return JSON.parse(b64urlDecode(payload));
+  } catch {
+    return null;
+  }
+}
+
+/** Check if token is valid: parseable, has email, not expired. */
+function isTokenValid(token: string): boolean {
+  const claims = parseToken(token);
+  if (!claims?.email) return false;
+  if (claims.exp && claims.exp * 1000 < Date.now()) return false;
+  return true;
+}
+
+/**
+ * Self-hosted auth gate. Shows login/signup when no token,
+ * renders children when authenticated.
+ * When VITE_AUTH_ENABLED is not "true", renders children directly (dev mode).
+ */
+export default function AuthGate({ children }: AuthGateProps) {
+  const authEnabled = import.meta.env.VITE_AUTH_ENABLED === "true";
+
+  // Initialize token AND token getter synchronously to avoid race condition
+  // where children's effects fire API calls before the getter is wired up.
+  const [token, setToken] = useState<string | null>(() => {
+    const t = localStorage.getItem("cthulu_auth_token");
+    if (t && isTokenValid(t)) {
+      setAuthTokenGetter(() => Promise.resolve(t));
+      return t;
+    }
+    // Clear invalid/expired token
+    if (t) localStorage.removeItem("cthulu_auth_token");
+    setAuthTokenGetter(null);
+    return null;
+  });
+
+  const logout = useCallback(() => {
+    localStorage.removeItem("cthulu_auth_token");
+    setAuthTokenGetter(null);
+    setToken(null);
+  }, []);
+
+  if (!authEnabled) return <>{children}</>;
+
+  if (token) {
+    const claims = parseToken(token);
+    return (
+      <AuthContext.Provider value={{ logout, userEmail: claims?.email ?? null }}>
+        {children}
+      </AuthContext.Provider>
+    );
+  }
+
+  return <AuthForm onSuccess={(t) => {
+    localStorage.setItem("cthulu_auth_token", t);
+    setAuthTokenGetter(() => Promise.resolve(t));
+    setToken(t);
+  }} />;
+}
+
+function AuthForm({ onSuccess }: { onSuccess: (token: string) => void }) {
+  const [mode, setMode] = useState<"login" | "signup">("login");
+  const [email, setEmail] = useState("");
+  const [password, setPassword] = useState("");
+  const [error, setError] = useState("");
+  const [loading, setLoading] = useState(false);
+
+  const handleSubmit = async (e: React.FormEvent) => {
+    e.preventDefault();
+    setError("");
+    setLoading(true);
+
+    try {
+      const endpoint = mode === "signup" ? "/api/auth/signup" : "/api/auth/login";
+      const res = await fetch(`${getServerUrl()}${endpoint}`, {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ email, password }),
+      });
+
+      const data = await res.json();
+
+      if (!res.ok) {
+        setError(data.error || `Failed (${res.status})`);
+        return;
+      }
+
+      onSuccess(data.token);
+    } catch (err) {
+      setError(err instanceof Error ? err.message : "Network error");
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  return (
+    <div className="auth-container">
+      <div className="auth-card">
+        <h1 className="auth-title">Cthulu</h1>
+        <p className="auth-subtitle">
+          {mode === "login" ? "Sign in to your account" : "Create a new account"}
+        </p>
+
+        <form onSubmit={handleSubmit} className="auth-form">
+          <input
+            type="email"
+            placeholder="Email"
+            value={email}
+            onChange={(e) => setEmail(e.target.value)}
+            className="auth-input"
+            required
+            autoFocus
+          />
+          <input
+            type="password"
+            placeholder="Password (8+ characters)"
+            value={password}
+            onChange={(e) => setPassword(e.target.value)}
+            className="auth-input"
+            required
+            minLength={8}
+          />
+
+          {error && <p className="auth-error">{error}</p>}
+
+          <button type="submit" className="auth-button" disabled={loading}>
+            {loading ? "..." : mode === "login" ? "Sign In" : "Sign Up"}
+          </button>
+        </form>
+
+        <button
+          className="auth-toggle"
+          onClick={() => { setMode(mode === "login" ? "signup" : "login"); setError(""); }}
+        >
+          {mode === "login"
+            ? "Don't have an account? Sign up"
+            : "Already have an account? Sign in"}
+        </button>
+      </div>
+    </div>
+  );
+}
--- a/cthulu-studio/src/components/Sidebar.tsx
+++ b/cthulu-studio/src/components/Sidebar.tsx
@@ -1,6 +1,6 @@
 import { useState, useEffect, useCallback } from "react";
 import { STUDIO_ASSISTANT_ID, type FlowSummary, type Flow, type NodeTypeSchema, type AgentSummary, type SavedPrompt, type ActiveView } from "../types/flow";
-import { listAgents, createAgent, deleteAgent, listPrompts, savePrompt, deletePrompt as deletePromptApi, listAgentSessions, newAgentSession } from "../api/client";
+import { listAgents, createAgent, deleteAgent, listPrompts, savePrompt, deletePrompt as deletePromptApi, listAgentSessions, newAgentSession, listTeams } from "../api/client";
 import type { InteractSessionInfo } from "../api/client";
 import { Switch } from "@/components/ui/switch";
 import { Collapsible, CollapsibleTrigger, CollapsibleContent } from "@/components/ui/collapsible";
@@ -66,6 +66,11 @@ export default function Sidebar({
   const [prompts, setPrompts] = useState<SavedPrompt[]>([]);
   const [agentMeta, setAgentMeta] = useState<Map<string, { busy: boolean; sessions: InteractSessionInfo[]; cost: number }>>(new Map());
   const [expandedAgents, setExpandedAgents] = useState<Set<string>>(new Set());
+  const [userTeams, setUserTeams] = useState<{ id: string; name: string }[]>([]);
+
+  useEffect(() => {
+    listTeams().then((res) => setUserTeams(res.teams.map((t) => ({ id: t.id, name: t.name })))).catch(() => {});
+  }, []);
 
   const refreshAgents = useCallback(async () => {
     try {
@@ -153,6 +158,20 @@ export default function Sidebar({
     }
   }
 
+  async function handleCreateTeamAgent(teamId: string) {
+    try {
+      const team = userTeams.find((t) => t.id === teamId);
+      const { id } = await createAgent({
+        name: `${team?.name || "Team"} Agent`,
+        team_id: teamId,
+      });
+      await refreshAgents();
+      onAgentCreated(id);
+    } catch (e) {
+      console.error("Failed to create team agent:", e);
+    }
+  }
+
   async function handleDeleteAgent(e: React.MouseEvent, agentId: string) {
     e.stopPropagation();
     if (!confirm("Delete this agent?")) return;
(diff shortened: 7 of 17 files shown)


saltyskip:

_Reviewing PR #122 — this may take a few minutes._

cthulu-backend/api/agents/chat.rs
```diff
@@ -792,6 +793,15 @@ pub(crate) async fn chat(
         (StatusCode::NOT_FOUND, Json(json!({ "error": "agent not found" })))
     })?;
 
+    // If team agent, verify sender is a team member
+    if let Some(ref team_id) = agent.team_id {
+        let ts = state.team_store.read().await;
+        if !ts.teams.get(team_id).map(|t| t.members.contains(&auth.user_id)).unwrap_or(false) {
+            return Err((StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))));
+        }
```
saltyskip, line 801: **Missing auth on reconnect endpoint**: The `stream_agent_chat` (reconnect) endpoint has no `AuthUser` extractor or team membership check. A non-member can reconnect to a team agent's SSE stream by guessing the agent ID + session ID, bypassing the team membership check you added here.

Add `auth: AuthUser` to `stream_agent_chat` and replicate this team membership check there too.

cthulu-backend/agents/mod.rs
```diff
@@ -96,6 +96,9 @@ pub struct Agent {
     /// If true, this agent is only usable as a sub-agent and is hidden from the UI agent list.
     #[serde(default)]
     pub subagent_only: bool,
+    /// When set, this agent belongs to a team. All team members can access it.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub team_id: Option<String>,
```
saltyskip, line 101: The `team_id` field on `Agent` is good. Make sure the reconnect endpoint (`stream_agent_chat` in chat.rs) also checks team membership — currently it doesn't have an `AuthUser` extractor, so this field is only checked on the initial `chat` call but not on reconnect.

cthulu-backend/api/agents/chat.rs
```diff
@@ -1440,6 +1452,16 @@ pub(crate) async fn chat(
...
         }
 
+        // Broadcast sender attribution for team session spectators
+        {
+            let evt = format!("user_message:{}", serde_json::to_string(&json!({
+                "user_id": sender_id_for_bc, "text": prompt_for_bc,
+            })).unwrap_or_default());
+            let _ = bc_tx.send(evt.clone());
+            let mut bufs = chat_event_buffers.lock().await;
+            if let Some(b) = bufs.get_mut(&proc_key_for_stream) { b.push(evt); }
```
saltyskip, line 1462: The `user_message` broadcast format works correctly with the `split_once(':')` SSE consumer pattern. One note: the `prompt_for_bc` is truncated to 200 chars (line 1223), but it's broadcast to all subscribers including the sender — the frontend should filter out the sender's own message to avoid duplication. Currently `useAgentChat.ts` renders it unconditionally.

saltyskip (changes requested):
## 🤖 Cthulu Review

### Blocking Issues

**1. TOCTOU race in signup** (`clerk_auth.rs`): The signup handler drops the read lock after checking if email exists, runs slow bcrypt hashing (~200ms at cost=12), then acquires the write lock to insert. A concurrent signup with the same email can create a duplicate user. Fix: hash the password first (no lock needed), then take the write lock for both check + insert.

**2. Non-atomic file writes** (`clerk_auth.rs:59`, `teams.rs:52`): Both `UserStore::save()` and `TeamStore::save()` use direct `std::fs::write()`. Per CLAUDE.md rule 9, persistence must use temp file + rename for crash safety:
```rust
std::fs::write(&tmp, &json)?;
std::fs::rename(&tmp, &path)
```

**3. Missing auth on reconnect endpoint** (`chat.rs`): `stream_agent_chat` has no `AuthUser` extractor or team membership check. A non-member can reconnect to a team agent's SSE stream by guessing agent ID + session ID, completely bypassing the team access control added to the `chat` handler.

**4. Side effect during render** (`AuthGate.tsx:62`): `setAuthTokenGetter()` is called directly in the component body. This runs on every render and violates CLAUDE.md rule 2. Move to `useEffect`:
```tsx
useEffect(() => {
  setAuthTokenGetter(() => Promise.resolve(token));
}, [token]);
```

### Non-blocking Concerns

- **Any member can remove any member** (`teams.rs:190`): Including the creator. A newly-added member could remove everyone else. Consider protecting the creator from removal.
- **Weak JWT secret**: `Uuid::new_v4()` gives 122 bits. Consider 256 bits from a CSPRNG for production.
- **Sender sees own user_message**: The `user_message` event is broadcast to all subscribers including the sender. The frontend `useAgentChat.ts` renders it unconditionally — the sender will see their own message echoed back as `[user_id]: text`.

### What Looks Good

- Team membership check on `chat` endpoint is clean and correct
- The `user_message` broadcast format works correctly with the `split_once(':')` SSE pattern
- The agent create picker UX (select dropdown) is a nice pattern
- Good test coverage on auth primitives (JWT roundtrip, bcrypt, store persistence)
- Auth is properly disabled by default in dev mode

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #122 (previous HEAD: `a65deff`, new HEAD: `f10fb18`)_

cthulu-backend/api/agents/chat.rs
```diff
@@ -792,6 +793,15 @@ pub(crate) async fn chat(
         (StatusCode::NOT_FOUND, Json(json!({ "error": "agent not found" })))
     })?;
 
+    // If team agent, verify sender is a team member
+    if let Some(ref team_id) = agent.team_id {
+        let ts = state.team_store.read().await;
+        if !ts.teams.get(team_id).map(|t| t.members.contains(&auth.user_id)).unwrap_or(false) {
+            return Err((StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))));
```
saltyskip, line 800: **Bug: `team_id` is never set on agents, so this check is dead code.**

`CreateAgentRequest` in `handlers.rs` has no `team_id` field, `AgentBuilder::build()` always sets `team_id: None`, and `list_agents` doesn't return `team_id`. The frontend sends `team_id` on creation but it's silently ignored by serde deserialization.

To fix:
1. Add `team_id: Option<String>` to `CreateAgentRequest`
2. Add a `.team_id()` method to `AgentBuilder`, or set it directly after `builder.build()`
3. Include `team_id` in the `list_agents` JSON response

Without these, team agents can never be created — this entire feature is non-functional.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
+        self.teams.values()
+            .filter(|t| t.members.contains(&user_id.to_string()))
```
saltyskip, line 50: **Non-atomic write — data loss risk on crash.**

`TeamStore::save()` uses `std::fs::write(path, json)` directly, which can corrupt the file if the process crashes mid-write. `UserStore::save()` correctly uses the temp-file-then-rename pattern. This should be consistent:

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
```
saltyskip, line 190: **Any member can remove any other member, including the creator.**

The doc says "no roles — all members are equal" but this means a newly-added member can immediately remove everyone else, effectively hijacking the team. Consider at minimum preventing removal of `team.created_by`, or requiring self-removal only (users can leave but can't kick others without a role system).

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,523 @@
...
+        Ok(secret) if !secret.trim().is_empty() => secret.trim().to_string(),
+        _ => {
+            // 256 bits of randomness: two UUID v4s concatenated (each has 122 bits)
+            let secret = format!("{}{}", uuid::Uuid::new_v4(), uuid::Uuid::new_v4())
+                .replace('-', "");
+            let _ = std::fs::write(&path, &secret);
+            secret
+        }
+    }
+}
```
saltyskip, line 120: **Weak JWT secret generation.** Two UUID v4s concatenated give ~244 bits of randomness (each has 122 random bits), but the string representation after removing dashes is hex-like, reducing effective entropy per character. Consider using `rand::thread_rng().gen::<[u8; 32]>()` encoded as hex or base64 for a cleaner 256-bit secret. Not a blocker, but worth noting.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,523 @@
...
+async fn search_users(
+    _auth: AuthUser,
+    axum::extract::State(state): axum::extract::State<AppState>,
+    axum::extract::Query(params): axum::extract::Query<std::collections::HashMap<String, String>>,
+) -> (StatusCode, Json<Value>) {
+    let query = params.get("q").map(|s| s.trim().to_lowercase()).unwrap_or_default();
+    if query.is_empty() {
+        return (StatusCode::OK, Json(json!({ "users": [] })));
+    }
+
```
saltyskip, line 457: **User search endpoint leaks user existence to any authenticated user.**

Any logged-in user can enumerate all emails/names in the system via this endpoint. This is needed for team member invites, but consider rate-limiting or requiring the caller to be a team owner/member before searching. At minimum, document this as an intentional trade-off.

cthulu-studio/src/components/chat/useAgentChat.ts
```diff
@@ -79,6 +79,15 @@ function applySSEEvent(
         setGitSnapshot(data as MultiRepoSnapshot);
       }
       return false;
+    } else if (event.type === "user_message") {
+      // Team session: another user sent a message — show attribution
+      const sender = data.user_id || "unknown";
+      const text = data.text || "";
```
saltyskip, line 85: **UX: Shows raw `user_id` (e.g. `abc123_def456`) instead of display name.**

The `user_message` event only carries `user_id` and `text`. Consider including the sender's display name or email in the broadcast event from the backend, so the chat shows `[Alice]: hello` instead of `[abc123_def456]: hello`.

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
     </Select>
   );
 }
+
+function ProfileMenu() {
+  const { logout, userEmail } = useAuth();
+  const [open, setOpen] = useState(false);
+  const [profile, setProfile] = useState<{ name: string | null; email: string } | null>(null);
```
saltyskip, line 295: **`loadData` has an empty dependency array but closes over nothing — this is fine.** However, the `useEffect` on line 303 will re-fetch profile + teams every time the dropdown opens. Consider caching or only fetching once, since the data rarely changes.

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,164 @@
...
+  useEffect(() => {
+    if (token) {
+      // Validate token is parseable; clear if corrupted
+      if (emailFromToken(token) === null) {
+        logout();
+        return;
+      }
+      setAuthTokenGetter(() => Promise.resolve(token));
+    } else {
+      setAuthTokenGetter(null);
```
saltyskip: **No handling for expired/revoked tokens.** `emailFromToken` checks parseability but not expiry. If a token expires, every API call will 401 but the user stays "logged in" on the client with no feedback. Consider adding a global response interceptor that calls `logout()` on 401, or checking `exp` in the token payload.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Blocking: Team agent creation is non-functional.** The `CreateAgentRequest` struct and `AgentBuilder` don't handle `team_id`, and `list_agents` doesn't return it. The frontend sends `team_id` on creation but it's silently dropped. This means:
- Team agents are created as personal agents (no `team_id` persisted)
- The sidebar team sections will always be empty
- The team membership check on `chat` is dead code

### Must fix before merge:
1. **Add `team_id` to `CreateAgentRequest`** and set it on the Agent in `create_agent` handler
2. **Include `team_id` in `list_agents` response** so the frontend can group agents by team
3. **Make `TeamStore::save()` atomic** (temp file + rename) — it currently uses bare `std::fs::write`, inconsistent with `UserStore::save()` which correctly does atomic writes

### Should fix:
4. **Broadcast display name, not `user_id`** in `user_message` SSE events — chat currently shows raw IDs like `[abc123_def456]`
5. **Handle expired tokens on the frontend** — expired JWTs cause silent 401s with no user feedback; add a global 401 interceptor that triggers logout
6. **Consider member removal authorization** — any member can remove any other member including the creator

### Minor observations:
- User search endpoint allows any authed user to enumerate all accounts (acceptable for team invites, but worth documenting)
- JWT secret generation using concatenated UUIDs works but isn't idiomatic — `rand` crate would be cleaner

The auth system, team CRUD, and UI components are well-structured. The core issue is just the missing backend wiring for `team_id` on agent creation and listing.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #122 (previous HEAD: `f10fb18`, new HEAD: `1ab1d23`)_

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
+        self.teams.values()
+            .filter(|t| t.members.contains(&user_id.to_string()))
+            .collect()
+    }
```
saltyskip, line 52: **Non-atomic save** — `TeamStore::save` uses `std::fs::write` directly, which can corrupt the file on crash. `UserStore::save` in `clerk_auth.rs` correctly uses temp-file + rename. This should follow the same pattern:

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,530 @@
+//! Self-hosted authentication: signup, login, JWT issuance + verification.
+//!
+//! Users stored in `{data_dir}/users.json`. Passwords hashed with bcrypt.
+//! JWTs signed with a server-generated secret stored in `{data_dir}/jwt_secret`.
+//!
+//! When `AUTH_ENABLED` env var is not "true" (default), auth is bypassed
+//! and all requests get a hardcoded "dev_user" identity.
```
saltyskip, line 7: **Misleading filename** — The file is named `clerk_auth.rs` but the module implements self-hosted authentication (email/password + bcrypt + JWT). This naming will confuse future contributors into thinking it uses Clerk. Consider renaming to `local_auth.rs` or `self_auth.rs`.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,530 @@
...
+        return (
+            StatusCode::BAD_REQUEST,
+            Json(json!({ "error": "Email required, password must be 8+ characters" })),
+        );
+    }
+
+    let email = body.email.trim().to_lowercase();
+    let password = body.password;
+
+    // Hash password outside lock (bcrypt is CPU-intensive)
```
saltyskip, line 287: **Minimal email validation** — Only checks `email.trim().is_empty()`. Users can sign up with emails like `foo` (no `@`). Consider at minimum:
```rust
if !email.contains('@') || email.len() < 5 {
    return (StatusCode::BAD_REQUEST, ...);
}
```

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,530 @@
...
+    match user {
+        Some(u) => (StatusCode::OK, Json(json!({
+            "id": u.id, "email": u.email, "name": u.name,
+            "avatar_url": u.avatar_url, "created_at": u.created_at,
+        }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
+    }
+}
+
+#[derive(Deserialize)]
```
saltyskip, line 411: **O(n) lookup by user ID** — `store.users.values().find(|u| u.id == auth.user_id)` is a linear scan because the store is keyed by email. Also appears in `update_profile`. Fine for small user counts, but consider a secondary index (id -> email map) if the user base grows.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
```
saltyskip, line 194: **Any member can remove any other member (including the creator)** — The `created_by` field exists on Team but isn't used for authorization. The doc says "no roles" which is intentional, but worth noting that any member can remove any other member, including the creator. If someone maliciously removes everyone, the team gets silently deleted (line 195-197).

cthulu-studio/src/components/chat/useAgentChat.ts
```diff
@@ -79,6 +79,15 @@ function applySSEEvent(
         setGitSnapshot(data as MultiRepoSnapshot);
       }
       return false;
+    } else if (event.type === "user_message") {
+      // Team session: another user sent a message — show attribution
+      const sender = data.user_id || "unknown";
```
saltyskip, line 84: **Verify SSE event type mapping** — The backend broadcasts `user_message:{json}` as a colon-delimited string via `bc_tx.send()`. This handler expects `event.type === "user_message"` with parsed `data.user_id` and `data.text`.

This only works if the SSE relay code (from the dependency PRs) splits the broadcast string on the first `:` to produce separate SSE `event:` / `data:` fields. If the relay sends the raw string as data instead, this handler will never fire. Please confirm the relay does the split.

saltyskip (changes requested):
## 🤖 Cthulu Review

Good feature design — team sessions with broadcast attribution is clean. Auth system is well-structured with proper bcrypt offloading and atomic user persistence. A few issues to address before merging:

### Must fix
- **`TeamStore::save()` is not atomic** — Uses `std::fs::write` directly, risking data corruption on crash. `UserStore::save` already does this correctly with temp+rename. Apply the same pattern.
- **Verify SSE relay compatibility** — Backend broadcasts `user_message:{json}` as a colon-delimited string. Frontend expects `event.type === "user_message"`. This only works if the SSE relay code (from dependency PRs) splits on the first `:`. If it doesn't, team message attribution will silently fail.

### Should fix
- **`clerk_auth.rs` naming is misleading** — Module implements self-hosted email/password auth, not Clerk. Will confuse contributors.
- **No email format validation on signup** — Only checks for empty string, allowing signups with `foo` (no `@`).

### Minor / worth noting
- `get_profile` / `update_profile` do O(n) scans by user ID when store is keyed by email. Fine for now.
- Any team member can remove any other member including the creator. Team auto-deletes when empty. Intentional per "no roles" design but worth documenting.
- `AuthContext.Provider` is absent when auth is disabled — `useAuth()` returns defaults. Works because `ProfileMenu` returns null, but future consumers should be aware.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #122 (previous HEAD: `1ab1d23`, new HEAD: `241f749`)_

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+            Ok(contents) => serde_json::from_str(&contents).unwrap_or_default(),
+            Err(_) => HashMap::new(),
+        };
+        Self { teams }
+    }
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
```
saltyskip, line 45: **Non-atomic write**: `TeamStore::save` uses direct `std::fs::write` instead of the temp-file + rename pattern. `UserStore::save` (in `local_auth.rs`) correctly does atomic writes via `.with_extension("json.tmp")` then `std::fs::rename`. A crash mid-write here could corrupt `teams.json`.

Suggested fix:
```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+            team.members.push(target_id);
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
+}
+
+async fn remove_member(
```
saltyskip, line 176: **Any team member can remove any other member, including the creator**: There's no distinction between the team creator and members for removal permissions. A newly invited member could immediately remove the team creator (and everyone else). Consider:
- Only allow self-removal (leaving the team), or
- Only allow the `created_by` user to remove others, or
- At minimum, prevent removing the `created_by` user

cthulu-backend/api/agents/chat.rs
```diff
@@ -792,6 +793,15 @@ pub(crate) async fn chat(
         (StatusCode::NOT_FOUND, Json(json!({ "error": "agent not found" })))
     })?;
 
+    // If team agent, verify sender is a team member
+    if let Some(ref team_id) = agent.team_id {
```
saltyskip, line 797: **Missing ownership check for personal agents**: The team membership check is correct for team agents, but there's no ownership check for personal agents (`team_id` is `None`). Any authenticated user could send messages to any other user's personal agent if they know the agent ID. This is a security concern for multi-user deployments.

Consider adding an ownership check when `team_id` is `None`.

cthulu-backend/api/agents/chat.rs
```diff
@@ -1440,6 +1452,16 @@ pub(crate) async fn chat(
             );
         }
 
+        // Broadcast sender attribution for team session spectators
+        {
+            let evt = format!("user_message:{}", serde_json::to_string(&json!({
```
saltyskip, line 1457: **Verify broadcast-to-SSE conversion**: The event is broadcast as `"user_message:{json}"` using a colon-delimited string. The frontend `applySSEEvent` expects `event.type === "user_message"`, which means the reconnect/spectate endpoint must parse this `"type:data"` format and convert it to proper SSE fields (`event: user_message\ndata: {json}`).

If this parsing already exists in the dependency PRs, this is fine. But if the reconnect path forwards the raw broadcast string as a generic SSE `data:` field, then `event.type` would never be `"user_message"` and this would silently fail. Worth verifying the end-to-end path.

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,164 @@
...
+    localStorage.removeItem("cthulu_auth_token");
+    setAuthTokenGetter(null);
+    setToken(null);
+  }, []);
+
+  // Sync token getter as side effect (not during render)
+  useEffect(() => {
+    if (token) {
+      // Validate token is parseable; clear if corrupted
+      if (emailFromToken(token) === null) {
```
saltyskip: **Token validation checks parseability but not expiry**: `emailFromToken(token) === null` catches corrupted tokens but not expired ones. With 24h JWT expiry, a user returning after a day will have a parseable but expired token — API calls will fail with 401 but the UI will render as if authenticated.

Suggested: also check the `exp` claim:
```typescript
function isTokenValid(token: string): boolean {
  try {
    const payload = token.split('.')[1];
    const json = atob(payload.replace(/-/g, '+').replace(/_/g, '/'));
    const { exp, email } = JSON.parse(json);
    return !!email && exp * 1000 > Date.now();
  } catch { return false; }
}
```

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+  const [teams, setTeams] = useState<{ id: string; name: string }[]>([]);
+  const [editingName, setEditingName] = useState(false);
+  const [nameInput, setNameInput] = useState("");
+  const [newTeamName, setNewTeamName] = useState("");
+  const [expandedTeam, setExpandedTeam] = useState<string | null>(null);
+  const [teamMembers, setTeamMembers] = useState<Record<string, { id: string; email: string; name: string | null }[]>>({});
+  const [memberError, setMemberError] = useState("");
+  const menuRef = useRef<HTMLDivElement>(null);
+
+  if (!userEmail) return null;
```
saltyskip, line 305: **Conditional return before hooks — React rules violation**: `if (!userEmail) return null` appears before `useCallback`, `useEffect`, and `useState` calls below it. React hooks must be called in the same order on every render — conditional early returns before hooks will crash when `userEmail` transitions between truthy/null.

Move this early return below all hook declarations.

cthulu-studio/src/components/Sidebar.tsx
```diff
@@ -66,6 +66,11 @@ export default function Sidebar({
   const [prompts, setPrompts] = useState<SavedPrompt[]>([]);
   const [agentMeta, setAgentMeta] = useState<Map<string, { busy: boolean; sessions: InteractSessionInfo[]; cost: number }>>(new Map());
   const [expandedAgents, setExpandedAgents] = useState<Set<string>>(new Set());
+  const [userTeams, setUserTeams] = useState<{ id: string; name: string }[]>([]);
+
+  useEffect(() => {
+    listTeams().then((res) => setUserTeams(res.teams.map((t) => ({ id: t.id, name: t.name })))).catch(() => {});
```
saltyskip, line 72: **Teams fetched once, never refreshed**: `listTeams()` is called in a `useEffect` with `[]` deps, so the team list is only fetched on mount. If a user creates a new team via the ProfileMenu dropdown, the sidebar won't show it until page reload. Consider lifting team state up or adding a refresh callback.

cthulu-studio/src/components/chat/useAgentChat.ts
```diff
@@ -79,6 +79,15 @@ function applySSEEvent(
         setGitSnapshot(data as MultiRepoSnapshot);
       }
       return false;
+    } else if (event.type === "user_message") {
```
saltyskip, line 82: **`user_message` displays raw user_id**: The sender attribution shows something like `[user_abc_123]: message text` which is the internal UUID-based user ID, not a human-readable name or email. Consider having the backend include the sender's display name in the broadcast payload:
```rust
"user_id": sender_id_for_bc, "display_name": sender_name, "text": prompt_for_bc,
```
Then render: `[display_name]: text`

saltyskip (changes requested):
## 🤖 Cthulu Review

Solid feature — team sessions, auth, and multi-user broadcast are well-structured. The auth module is particularly clean (atomic saves, bcrypt off-thread, CSPRNG secrets, proper file permissions). However, there are a few issues that should be addressed before merge:

### Blocking

1. **React hooks rules violation in `ProfileMenu`** (`TopBar.tsx:305`): `if (!userEmail) return null` appears before hooks (`useCallback`, `useEffect`, `useState`). This will crash when `userEmail` transitions. Move the early return below all hook declarations.

2. **Non-atomic `TeamStore::save`** (`teams.rs:45`): Uses direct `std::fs::write` while `UserStore::save` correctly uses temp+rename. Inconsistent and crash-unsafe — easy fix.

3. **No personal agent ownership check** (`chat.rs:797`): Team agents correctly verify membership, but personal agents (`team_id: None`) have no ownership check. Any authenticated user can chat with any agent by ID.

### Non-blocking (but worth addressing)

4. **Any member can remove any other member** including the team creator (`teams.rs:176`). Consider restricting removal permissions.

5. **Token expiry not checked client-side** (`AuthGate.tsx:57`): Expired JWTs will pass the parseable check, causing silent 401 failures.

6. **`user_message` broadcast format** (`chat.rs:1457`): Verify the reconnect endpoint correctly splits `"user_message:{json}"` into SSE event type + data fields. If it forwards raw, the frontend handler will never fire.

7. **User ID shown instead of display name** in team chat attribution (`useAgentChat.ts:82`).

8. **Sidebar team list never refreshes** after creating a team in ProfileMenu (`Sidebar.tsx:72`).

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #122 (previous HEAD: `241f749`, new HEAD: `d7d5f04`)_

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
+        self.teams.values()
+            .filter(|t| t.members.contains(&user_id.to_string()))
+            .collect()
```
saltyskip, line 51: **Bug: Non-atomic save** — `TeamStore::save` uses direct `std::fs::write`, but `UserStore::save` in `local_auth.rs` correctly uses temp file + rename (crash-safe). A crash or power loss mid-write here will corrupt `teams.json`.

This violates the project's atomic persistence rule. Fix:

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
```
saltyskip, line 190: **Design: Any member can remove any other member, including the team creator.** There's no permission check — member A can kick member B (and even the creator). This seems like it should at least require `created_by == auth.user_id` or some admin concept, otherwise any member can empty the team.

At minimum, consider preventing removal of the `created_by` user by other members.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,568 @@
...
+            Json(json!({ "error": "Valid email required" })),
+        );
+    }
+    if body.password.len() < 8 {
+        return (
+            StatusCode::BAD_REQUEST,
+            Json(json!({ "error": "Password must be at least 8 bytes" })),
+        );
+    }
+    let password = body.password;
```
saltyskip, line 285: **Security: No max password length** — bcrypt silently truncates input at 72 bytes. A user who sets a 100-char password will authenticate with just the first 72 bytes, which is surprising and could be a security issue. Consider adding:

```rust
if body.password.len() > 72 {
    return (StatusCode::BAD_REQUEST, Json(json!({ "error": "Password must be at most 72 bytes" })));
}
```

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,568 @@
...
+    axum::extract::State(state): axum::extract::State<AppState>,
+    axum::extract::Query(params): axum::extract::Query<std::collections::HashMap<String, String>>,
+) -> (StatusCode, Json<Value>) {
+    let query = params.get("q").map(|s| s.trim().to_lowercase()).unwrap_or_default();
+    if query.is_empty() {
+        return (StatusCode::OK, Json(json!({ "users": [] })));
+    }
+
+    let store = state.user_store.read().await;
+    let results: Vec<Value> = store.users.values()
```
saltyskip, line 459: **Security: User enumeration via search** — The `/auth/users/search` endpoint lets any authenticated user search all users by email substring with no rate limiting. This enables full user-email enumeration. Consider:
- Only returning users who share a team with the requester
- Or adding a minimum query length > 1 char
- Or rate limiting this endpoint

cthulu-studio/src/components/TopBar.tsx
saltyskip, line 305: **Bug: Rules of Hooks violation** — `if (!userEmail) return null` is placed *before* `useCallback` and `useEffect` calls below. React requires hooks to be called in the same order on every render. When `userEmail` is null, the early return skips those hooks, which will cause warnings and potential crashes.

Move this early return *after* all hook declarations:
```tsx
const loadData = useCallback(async () => { ... }, []);
useEffect(() => { if (open) loadData(); }, [open, loadData]);
useEffect(() => { ... }, [open]);
// ... all hooks above
if (!userEmail) return null;
```

saltyskip (changes requested):
## 🤖 Cthulu Review

Solid feature — the auth system, team model, and frontend integration are well-structured. A few issues to address before merging:

### Blocking

1. **`TeamStore::save` is not atomic** — Uses `std::fs::write` directly instead of temp file + rename. `UserStore::save` does this correctly. A crash mid-write corrupts `teams.json`. See inline comment.

2. **Rules of Hooks violation in `ProfileMenu`** — `if (!userEmail) return null` appears before `useCallback`/`useEffect` declarations. React will crash or warn when `userEmail` transitions between null and non-null. Move the guard after all hooks.

3. **bcrypt 72-byte truncation** — No upper bound on password length. bcrypt silently truncates at 72 bytes, so a user with a 100-char password can authenticate with just the first 72 chars. Add a max length check or document the behavior.

### Non-blocking

4. **Any team member can remove any other member** (including the creator). Consider restricting removal to the creator or adding an admin concept.

5. **User search endpoint enables email enumeration** — Any authenticated user can search all emails with no rate limiting. Consider scoping results to teammates.

6. **Team agents in sidebar are missing delete button and session expansion** — Personal agents have these affordances but team agents don't. Intentional deferral?

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #122 (previous HEAD: `d7d5f04`, new HEAD: `737a5b5`)_

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
+        self.teams.values()
+            .filter(|t| t.members.contains(&user_id.to_string()))
+            .collect()
+    }
+}
+
```
saltyskip, line 54: **Bug: Non-atomic write — data corruption risk on crash.**

`TeamStore::save()` writes directly with `std::fs::write` instead of using the temp-file + rename pattern. Compare with `UserStore::save()` in `local_auth.rs` (same PR) which correctly uses atomic writes. If the process crashes mid-write, `teams.json` will be truncated/corrupt.

CLAUDE.md rule #9 prescribes atomic persistence via temp + rename. Fix:

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/agents/mod.rs
saltyskip, line 101: **Security: No server-side validation of `team_id` on agent creation.**

The `team_id` field is accepted from the client, but the agent creation endpoint doesn't appear to verify that the requesting user is a member of the specified team. A user could create agents under any team by guessing/knowing a `team_id`.

The chat endpoint correctly validates team membership (chat.rs L798), but the same check is needed at creation time. The create handler should verify `team_store.teams.get(team_id).members.contains(auth.user_id)` before persisting.

cthulu-backend/api/agents/chat.rs
```diff
@@ -1440,6 +1452,16 @@ pub(crate) async fn chat(
...
+                "user_id": sender_id_for_bc, "text": prompt_for_bc,
+            })).unwrap_or_default());
+            let _ = bc_tx.send(evt.clone());
+            let mut bufs = chat_event_buffers.lock().await;
+            if let Some(b) = bufs.get_mut(&proc_key_for_stream) { b.push(evt); }
+        }
```
saltyskip, line 1463: **Minor: `user_message` broadcast format uses colon-separated `type:data` string.**

The broadcast sends `"user_message:{json}"` as a raw string. On the frontend (`useAgentChat.ts`), the SSE parsing presumably splits this to get `event.type === "user_message"`. This works, but differs from the standard SSE `Event::default().event("user_message").data(json)` pattern used elsewhere in the file.

Worth verifying the SSE dispatch code that reads from `bc_tx` correctly parses this format into separate `event` and `data` fields for the client.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+            team.members.retain(|m| m != &user_id);
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
+        }
+        Some(_) => (StatusCode::FORBIDDEN, Json(json!({ "error": "Not a member of this team" }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "Team not found" }))),
+    }
```
saltyskip, line 193: **Design: Any team member can remove any other member (including the creator).**

The `remove_member` handler only checks that the requesting user is *a* member — not the team owner or the target user. This means:
- User A creates a team, adds User B
- User B can remove User A
- If User A was the last member, the team is deleted

This may be intentional per the "no roles — all members are equal" design, but it's worth confirming. At minimum, consider preventing removal of the `created_by` user, or adding a confirmation note in the API docs.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,566 @@
...
+    exp: usize,
+}
+
+const JWT_EXPIRY_HOURS: i64 = 24;
+
+/// Load or generate the JWT signing secret (256-bit CSPRNG, base64url-encoded).
+/// The secret file is created with mode 0600 on Unix.
+/// If the file is deleted, a new secret is generated and all existing JWTs are invalidated.
+pub fn load_or_create_jwt_secret(data_dir: &PathBuf) -> String {
+    let path = data_dir.join("jwt_secret");
```
saltyskip, line 100: **Nit: 24-hour JWT expiry with no refresh mechanism.**

JWTs expire after 24 hours (`JWT_EXPIRY_HOURS: i64 = 24`), but there's no refresh token endpoint. Users will be silently logged out and see 401 errors. The `AuthGate` client-side validation only checks parseability, not expiry, so the UI will show the logged-in state until the next API call fails.

Consider either: (a) adding a `/auth/refresh` endpoint, (b) extending the expiry (e.g. 7 days), or (c) having the frontend detect 401 responses and auto-logout.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,566 @@
...
+async fn signup(
+    axum::extract::State(state): axum::extract::State<AppState>,
+    Json(body): Json<SignupRequest>,
+) -> (StatusCode, Json<Value>) {
+    let email = body.email.trim().to_lowercase();
+    let email_parts: Vec<&str> = email.split('@').collect();
+    if email_parts.len() != 2 || email_parts[0].is_empty() || email_parts[1].is_empty() || !email_parts[1].contains('.') {
+        return (
+            StatusCode::BAD_REQUEST,
+            Json(json!({ "error": "Valid email required" })),
```
saltyskip, line 276: **Security: No rate limiting on signup/login endpoints.**

Bcrypt cost factor 12 provides ~250ms per attempt, which limits brute force somewhat, but there's no per-IP or per-email rate limit. An attacker could hammer `/auth/login` at ~4 req/s per core. For a self-hosted tool this may be acceptable risk, but worth adding a TODO or noting as a known limitation.

cthulu-studio/src/components/Sidebar.tsx
```diff
@@ -66,6 +66,11 @@ export default function Sidebar({
   const [prompts, setPrompts] = useState<SavedPrompt[]>([]);
   const [agentMeta, setAgentMeta] = useState<Map<string, { busy: boolean; sessions: InteractSessionInfo[]; cost: number }>>(new Map());
   const [expandedAgents, setExpandedAgents] = useState<Set<string>>(new Set());
+  const [userTeams, setUserTeams] = useState<{ id: string; name: string }[]>([]);
+
+  useEffect(() => {
```
saltyskip, line 71: **Bug: Teams are fetched once on mount but never refreshed.**

The `useEffect` fetches `listTeams()` on mount with `[]` deps, but if the user creates a new team (via ProfileMenu in TopBar), the sidebar won't reflect it. The `refreshAgents` callback is called after creating team agents, but there's no corresponding `refreshTeams` that re-fetches the teams list.

Consider either: (a) lifting team state up so both TopBar and Sidebar share it, or (b) adding a refresh mechanism similar to `refreshAgents`.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Requesting changes** — two issues need fixing before merge:

### Must fix

1. **`TeamStore::save()` is not atomic** — uses `std::fs::write` directly instead of temp+rename. Crash during write = corrupt `teams.json`. The `UserStore::save()` in the same PR does it correctly. One-line fix.

2. **No team membership validation on agent creation** — the chat endpoint correctly checks `team_store.teams.get(team_id).members.contains(user_id)`, but the agent creation endpoint accepts any `team_id` from the client without verification. A user can attach agents to teams they don't belong to.

### Should fix

3. **Sidebar teams never refresh** — fetched once on mount (`[]` deps), but creating a team in ProfileMenu won't update the sidebar. Needs a shared refresh mechanism.

4. **`user_message` broadcast format** — uses raw `"user_message:{json}"` string concatenation, which diverges from the standard SSE `Event::default().event().data()` pattern. Verify the SSE dispatch layer correctly parses this into separate event type + data for the frontend.

### Worth noting (non-blocking)

- **24h JWT with no refresh** — users get silently logged out. Consider a refresh endpoint or longer expiry.
- **No rate limiting on auth endpoints** — acceptable for self-hosted, but worth a TODO.
- **Any member can remove any member** — intentional per "no roles" design, but the creator can be kicked from their own team.
- **PR scope** — description says "~146 lines, 7 files" but actual diff is 1810 lines across 20 files (includes auth infra from dependency chain). Consider updating the description for reviewer context.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #122 (previous HEAD: `737a5b5`, new HEAD: `b81f09e`)_

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
+        self.teams.values()
+            .filter(|t| t.members.contains(&user_id.to_string()))
+            .collect()
+    }
```
saltyskip, line 52: **Non-atomic save — violates project convention #9.**

`UserStore::save()` correctly uses temp-file + rename for crash safety, but `TeamStore::save()` does a direct `std::fs::write`. If the process crashes mid-write, `teams.json` could be corrupted/truncated.

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/agents/chat.rs
```diff
@@ -792,6 +793,15 @@ pub(crate) async fn chat(
         (StatusCode::NOT_FOUND, Json(json!({ "error": "agent not found" })))
     })?;
 
+    // If team agent, verify sender is a team member
+    if let Some(ref team_id) = agent.team_id {
+        let ts = state.team_store.read().await;
```
saltyskip, line 798: **Team membership check only on `chat` — other agent endpoints are unguarded.**

This checks team membership for the chat/SSE endpoint, but what about agent deletion, update, and other CRUD operations? A non-member could potentially delete or modify a team agent through the existing agent endpoints.

Consider adding the same team membership guard to the agent update/delete handlers, or extracting a shared `verify_team_access()` helper.

cthulu-backend/api/agents/chat.rs
```diff
@@ -1440,6 +1452,16 @@ pub(crate) async fn chat(
             );
         }
 
+        // Broadcast sender attribution for team session spectators
+        {
+            let evt = format!("user_message:{}", serde_json::to_string(&json!({
+                "user_id": sender_id_for_bc, "text": prompt_for_bc,
+            })).unwrap_or_default());
+            let _ = bc_tx.send(evt.clone());
```
saltyskip, line 1460: **`user_message` event is broadcast unconditionally — even for personal agents.**

This broadcasts the sender attribution event for every chat message, regardless of whether the agent is team-owned. For personal agents (no `team_id`), spectators would see unnecessary `user_message` events. Consider gating this on `agent.team_id.is_some()`.

```rust
if agent.team_id.is_some() {
    let evt = format!("user_message:{}", ...);
    let _ = bc_tx.send(evt.clone());
    // ...
}
```

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,569 @@
...
+    axum::extract::State(state): axum::extract::State<AppState>,
+) -> (StatusCode, Json<Value>) {
+    let store = state.user_store.read().await;
+    let user = store.users.values().find(|u| u.id == auth.user_id);
+    match user {
+        Some(u) => (StatusCode::OK, Json(json!({
+            "id": u.id, "email": u.email, "name": u.name,
+            "avatar_url": u.avatar_url, "created_at": u.created_at,
+        }))),
+        None => (StatusCode::NOT_FOUND, Json(json!({ "error": "User not found" }))),
```
saltyskip, line 407: **O(n) scan to find user by ID — HashMap is keyed by email.**

Both `get_profile` and `update_profile` do `store.users.values().find(|u| u.id == auth.user_id)` which is a linear scan. Consider adding a secondary index (`HashMap<String, String>` mapping `id -> email`) if this becomes a hot path, or switching the primary key to user ID.

cthulu-backend/api/teams.rs
saltyskip, line 190: **Any member can remove any other member — including the team creator.**

There's no ownership/admin check. A newly-added member could immediately remove the creator and all other members. At minimum, consider preventing removal of the `created_by` user, or requiring `auth.user_id == created_by` for member removal of others.

cthulu-studio/src/components/Sidebar.tsx
```diff
@@ -319,6 +347,49 @@ export default function Sidebar({
...
+                {teamAgents.length === 0 && <div className="sidebar-empty">No agents yet</div>}
+                {teamAgents.map((agent) => {
+                  const meta = agentMeta.get(agent.id);
+                  const isActive = agent.id === selectedAgentId && activeView === "agent-workspace";
+                  return (
+                    <div key={agent.id} className="sb-agent">
+                      <div className={`sb-agent-row sb-agent-team${isActive ? " sb-agent-active" : ""}`}
+                        onClick={() => {
+                          const sessions = meta?.sessions ?? [];
+                          if (sessions.length > 0) onSelectSession(agent.id, sessions[0].session_id);
```
saltyskip, line 378: **Team agents have no delete button and no session-create on click.**

Personal agents get delete buttons and session management, but team agents only respond to click if `sessions.length > 0`. Clicking a freshly created team agent does nothing — users can't start a session or delete it. This seems like an incomplete UX.

Consider mirroring the personal agent UX or at minimum creating a new session on click when none exist.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,569 @@
...
+    /// `Json(user)` or `serde_json::to_value(user)` in HTTP responses.
+    pub password_hash: String,
+    pub name: Option<String>,
+    pub avatar_url: Option<String>,
+    pub created_at: String,
+}
+
+/// In-memory user store backed by `{data_dir}/users.json`.
+pub struct UserStore {
+    pub users: HashMap<String, StoredUser>, // email -> StoredUser
```
saltyskip, line 43: **`password_hash` is publicly serializable — high-risk footgun.**

The WARNING comment is helpful, but since `StoredUser` derives both `Serialize` and `Clone`, any accidental `Json(user.clone())` in a future endpoint will leak the bcrypt hash. Consider splitting into two structs (one for persistence with hash, one for API responses without), or using `#[serde(skip_serializing)]` on `password_hash` and a separate `fn to_json_for_disk()` method for file persistence.

saltyskip (changes requested):
## 🤖 Cthulu Review

Good overall structure — the auth system is well-designed (constant-time login, atomic user persistence, JWT with CSPRNG secret, bcrypt offloaded to spawn_blocking). The multi-tenant data isolation in `user_context.rs` is clean and well-tested.

**Must fix before merge:**

1. **`TeamStore::save()` is not atomic** — uses direct `std::fs::write` instead of temp-file + rename. This violates project convention #9 and risks corrupting `teams.json` on crash. `UserStore::save()` already does this correctly — just copy the pattern.

2. **Team membership not checked on non-chat agent endpoints** — the team access guard only protects the chat/SSE endpoint. Agent delete, update, and other CRUD operations are unguarded, allowing any user to modify/delete team agents they don't belong to.

3. **`user_message` broadcast fires for personal agents too** — the sender attribution event is emitted unconditionally. Should be gated on `agent.team_id.is_some()`.

**Should fix:**

4. **Any team member can remove any other member including the creator** — no ownership/admin protection. At minimum, prevent removal of `created_by` user.

5. **Team agents in sidebar are not fully functional** — no delete button, clicking does nothing when no sessions exist. Incomplete UX compared to personal agents.

6. **`password_hash` on a `Serialize + Clone` struct is a footgun** — the WARNING comment helps but a structural fix (separate persistence DTO or `#[serde(skip_serializing)]` + manual disk serialization) would be safer.

---

_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #122 (previous HEAD: `b81f09e`, new HEAD: `bad535d`)_

cthulu-backend/api/teams.rs
saltyskip, line 51: **Bug: Non-atomic save — data loss risk on crash.**

`UserStore::save` correctly uses temp-file + rename for crash safety, but `TeamStore::save` writes directly to `teams.json`. If the process crashes mid-write, you get a truncated/corrupt file.

This also violates CLAUDE.md rule #9 ("Atomic YAML persistence: Always write via temp file + rename").

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/teams.rs
saltyskip, line 190: **Security: Any team member can remove any other member.**

The `remove_member` handler only checks that the requesting user is *a* member of the team, not that they have any elevated role. This means any member can remove any other member (including the creator). Combined with the fact that `remove_member` auto-deletes the team when empty (line 191-193), a single malicious member could remove everyone else and destroy the team.

Consider either:
1. Only allowing the creator (`team.created_by`) to remove others
2. Allowing self-removal but restricting removal of others
3. At minimum, preventing removal of the `created_by` user

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,569 @@
...
+
+    // Constant-time: always run bcrypt verify to prevent user enumeration via timing.
+    // If user doesn't exist, verify against a dummy hash (same cost as real verify).
+    let dummy_hash = "$2b$12$000000000000000000000u2a5OJr0FkDxcMkGCuaLxMPqOIZJcMK";
+    let password = body.password.clone();
+    let hash = user.as_ref().map(|u| u.password_hash.clone())
+        .unwrap_or_else(|| dummy_hash.to_string());
+    let valid = tokio::task::spawn_blocking(move || bcrypt::verify(&password, &hash).unwrap_or(false))
+        .await
+        .unwrap_or(false);
```
saltyskip, line 365: **Concern: Hardcoded dummy bcrypt hash may not have the right cost factor.**

The dummy hash uses cost 12 prefix (`$2b$12$`) which matches signup, so the timing defense should work. However, if the cost factor in `signup` ever changes, this hardcoded string needs to be updated too. Consider generating it once at startup instead:

```rust
let dummy_hash = bcrypt::hash("dummy", 12).unwrap();
```

Or at least add a comment linking the two.

cthulu-backend/api/local_auth.rs
```diff
@@ -0,0 +1,569 @@
...
+    }
+
+    /// Atomic save: write to temp file, then rename (crash-safe).
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("users.json");
+        let tmp = path.with_extension("json.tmp");
+        let json = serde_json::to_string_pretty(&self.users)?;
+        std::fs::write(&tmp, json)?;
+        std::fs::rename(&tmp, &path)
+    }
```
saltyskip, line 74: **Consistency: `UserStore::save` uses `&PathBuf` while `load` also uses `&PathBuf`.**

Minor: idiomatic Rust would take `&Path` instead of `&PathBuf` (same applies to `TeamStore`). Not blocking, but worth a follow-up.

cthulu-backend/api/agents/chat.rs
saltyskip, line 1462: **Race condition: `user_message` broadcast happens before the agent processes the message.**

The `user_message` event is broadcast (lines 1455-1461) before the background reader task spawns and the agent even starts processing. If a spectator sees `[alice]: summarize this` but the agent hasn't started yet, the UX could be confusing — the user message appears but nothing follows for a while.

This is minor since the agent stream events will follow, but worth noting if timing matters for the UX.

cthulu-backend/api/local_auth.rs
saltyskip, line 407: **Performance: `get_profile` does a linear scan over all users to find by ID.**

`store.users` is keyed by email, but this does `users.values().find(|u| u.id == auth.user_id)`. Same issue in `update_profile`. With many users this is O(n) per request.

Consider adding a secondary `id -> email` index, or at minimum add the `email` to JWT claims so you can look up by email directly (you already store `email` in `Claims`).

cthulu-studio/src/api/client.ts
```diff
@@ -47,18 +80,33 @@ async function apiFetch<T>(
...
         "Content-Type": "application/json",
+        ...authHeaders,
         ...options.headers,
       },
     });
 
     const elapsed = Math.round(performance.now() - start);
 
     if (!res.ok) {
+      // Auto-logout on 401 (expired/invalid token)
```
saltyskip, line 105: **UX: `window.location.reload()` on 401 is aggressive.**

This fires on ANY 401, even if it's a one-off race condition or if the user is in the middle of typing. It also happens before the `throw new Error(...)` on the next line, so the reload races with error handling.

Consider:
1. Only reloading if the token was actually present (it's already checked with `getTokenFn`)
2. Clearing the token and letting the React state drive the transition back to the login form (via AuthGate's state), instead of a hard page reload

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+
+  const loadData = useCallback(async () => {
+    try {
+      const [p, t] = await Promise.all([api.getProfile(), api.listTeams()]);
+      setProfile({ name: p.name, email: p.email });
+      setTeams(t.teams.map((tm) => ({ id: tm.id, name: tm.name })));
+    } catch { /* ignore if not authed */ }
```
saltyskip, line 312: **Hook rules violation: `useCallback`/`useEffect` called conditionally.**

The `loadData` callback is defined with `useCallback` after the early return `if (!userEmail) return null;` on line 308. React hooks must not be called conditionally — this will crash or produce incorrect behavior when `userEmail` transitions between null and a value.

Move the early return after all hook declarations, or restructure as:
```tsx
function ProfileMenu() {
  const { logout, userEmail } = useAuth();
  const [open, setOpen] = useState(false);
  // ... all hooks here ...
  
  if (!userEmail) return null;
  // ... render ...
}
```

cthulu-studio/src/components/chat/useAgentChat.ts
```diff
@@ -79,6 +79,15 @@ function applySSEEvent(
...
+      // Team session: another user sent a message — show attribution
+      const sender = data.user_id || "unknown";
+      const text = data.text || "";
+      partsRef.current = [...partsRef.current, {
+        type: "text",
+        text: `[${sender}]: ${text}`,
```
saltyskip, line 88: **Minor: `user_message` text is displayed raw without sanitization.**

The `text` field is rendered as `[${sender}]: ${text}` which is fine for now since React escapes HTML in text nodes. But the `sender` (user_id) and `text` come from the broadcast and are not validated on the frontend. If the backend ever passes through unsanitized content, this could be a vector. Currently safe because React handles escaping.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Scope**: This PR is much larger than described (22 files, ~1850 lines vs. claimed 7 files, ~146 lines). It includes a full auth system (signup/login/JWT), teams CRUD, user context/multi-tenancy, profile management, AuthGate UI, and team sessions. The PR description should be updated to reflect the actual scope.

### Blocking Issues

1. **React hooks called conditionally** (`TopBar.tsx:308-312`): `ProfileMenu` has an early `return null` before `useCallback`/`useEffect` declarations. This violates React's rules of hooks and will cause crashes or bugs when `userEmail` transitions. Must fix.

2. **TeamStore uses non-atomic save** (`teams.rs:48-52`): Direct `fs::write` to `teams.json` risks data corruption on crash. `UserStore` correctly uses temp-file + rename. This violates the project's own rule #9.

3. **Any team member can remove any other member** (`teams.rs:182-200`): No authorization beyond "is a member" — any member can kick the creator and destroy the team. Needs at minimum a `created_by` check or self-removal-only policy.

### Non-blocking Observations

- `get_profile` / `update_profile` do O(n) linear scans over users (keyed by email, searched by ID). The email is already in JWT claims — could look up directly.
- `window.location.reload()` on 401 is aggressive — consider React state-driven logout instead.
- Dummy bcrypt hash in `login` is hardcoded — will diverge if cost factor changes.
- PR description says "7 files, ~146 lines" but actual diff is 22 files, ~1850 lines. Misleading for reviewers.

### What looks good

- Constant-time bcrypt verify for login (anti-enumeration)
- JWT secret generation with CSPRNG + 0600 permissions
- Auth token extraction supports both header and query param (for SSE)
- User ID charset validation in the extractor prevents path traversal
- Atomic save on UserStore
- Good test coverage for auth roundtrips and store persistence

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #122 (previous HEAD: `bad535d`, new HEAD: `dc5084a`)_

cthulu-backend/agents/mod.rs
```diff
@@ -214,6 +217,7 @@ impl AgentBuilder<Ready> {
             hooks: self.hooks,
             subagents: self.subagents,
             subagent_only: self.subagent_only,
+            team_id: None,
```
saltyskip, line 220: **Bug: `team_id` is hardcoded to `None` in the builder, and `CreateAgentRequest` in `handlers.rs` doesn't include a `team_id` field.** The frontend sends `team_id` via `createAgent()`, but it gets silently dropped by serde (no `deny_unknown_fields`). Team agents can never actually be created.

Fix: add `team_id: Option<String>` to `CreateAgentRequest`, pass it through the builder or set it directly on the built agent:
```rust
let mut agent = builder.build();
agent.team_id = body.team_id;
```

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    pub fn save(&self, data_dir: &PathBuf) -> std::io::Result<()> {
+        let path = data_dir.join("teams.json");
+        let json = serde_json::to_string_pretty(&self.teams)?;
+        std::fs::write(path, json)
+    }
+
+    pub fn teams_for_user(&self, user_id: &str) -> Vec<&Team> {
+        self.teams.values()
```
saltyskip, line 49: **`TeamStore::save()` uses direct `std::fs::write` instead of atomic temp+rename.** `UserStore::save()` in the same codebase correctly does temp file + rename for crash safety. A crash mid-write here corrupts `teams.json`.

```rust
    std::fs::write(&tmp, json)?;
    std::fs::rename(&tmp, &path)
}
```

cthulu-backend/api/agents/chat.rs
saltyskip, line 797: The team membership check is only on the `chat` endpoint. The `delete_agent` and `update_agent` handlers in `handlers.rs` have no team checks — any authenticated user can delete or modify a team agent they don't belong to. At minimum, `delete_agent` should verify the caller is a team member (or the creator) before allowing deletion.

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+) -> (StatusCode, Json<Value>) {
+    let mut store = state.team_store.write().await;
+    match store.teams.get_mut(&id) {
+        Some(team) if team.members.contains(&auth.user_id) => {
+            team.members.retain(|m| m != &user_id);
+            if team.members.is_empty() {
+                store.teams.remove(&id);
+            }
+            let _ = store.save(&state.data_dir);
+            (StatusCode::OK, Json(json!({ "ok": true })))
```
saltyskip, line 189: **Any member can remove any other member, including the team creator.** Since teams have "no roles," a newly-added member can immediately remove the person who created the team. Consider either:
- Preventing removal of `created_by`
- Only allowing `created_by` to remove others
- Or at least documenting this as intentional

cthulu-backend/api/teams.rs
```diff
@@ -0,0 +1,240 @@
...
+    let mut store = state.team_store.write().await;
+    let response = json!({ "team": &team });
+    store.teams.insert(team.id.clone(), team);
+    let _ = store.save(&state.data_dir);
+
+    (StatusCode::CREATED, Json(response))
+}
+
+async fn get_team(
+    auth: AuthUser,
```
saltyskip, line 108: Silently ignoring save errors (`let _ = store.save(...)`) means the in-memory state diverges from disk on write failure. On next restart, the team is lost. This pattern appears in all handlers. Consider at least logging the error:
```rust
    tracing::error!(error = %e, "failed to persist team store");
}
```

cthulu-studio/src/components/TopBar.tsx
```diff
@@ -285,3 +288,186 @@ function ThemeSelector() {
...
+        value={query} onChange={(e) => handleChange(e.target.value)} />
+      {(results.length > 0 || searching) && (
+        <div className="user-search-results">
+          {searching && <div className="user-search-item user-search-loading">Searching...</div>}
+          {results.map((u) => (
+            <button key={u.id} className="user-search-item"
+              onClick={() => { onSelect(u.id, u.email); setQuery(""); setResults([]); }}>
+              <span className="user-search-name">{u.name || u.email.split("@")[0]}</span>
+              <span className="user-search-email">{u.email}</span>
+            </button>
```
saltyskip, line 467: The `UserSearchInput` debounce `setTimeout` is stored in `debounceRef` but never cleaned up on unmount. If the component unmounts mid-debounce, the callback fires on a stale component. Add:
```tsx
useEffect(() => {
  return () => { if (debounceRef.current) clearTimeout(debounceRef.current); };
}, []);
```

cthulu-studio/src/components/chat/useAgentChat.ts
saltyskip, line 85: The `user_id` shown here is the internal UUID (e.g., `[a1b2c3d4_...]`), not a human-readable name or email. Consider resolving the display name on the backend when constructing the `user_message` event, or at least using the email from the JWT claims instead of `auth.user_id`.

cthulu-backend/api/local_auth.rs
```diff
@@ -7,7 +7,7 @@
 //! and all requests get a hardcoded "dev_user" identity.
```
saltyskip, line 7: Nit: `put` is imported from `axum::routing` but unused — the `.put(update_profile)` on line 263 is a method on `MethodRouter`, not the free function.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Blocking: `team_id` never reaches the backend.** The `CreateAgentRequest` struct in `handlers.rs` doesn't include `team_id`, and `AgentBuilder::build()` hardcodes it to `None`. The frontend sends it, serde silently drops it. Team agents cannot be created — the core feature of this PR is broken.

### Other issues requiring fixes

1. **`TeamStore::save()` is not atomic** — uses `std::fs::write` directly instead of temp+rename. `UserStore` in the same codebase does this correctly. A crash mid-write corrupts `teams.json`.

2. **No team auth on agent delete/update** — `chat` checks team membership, but `delete_agent` and `update_agent` have no team checks. Any authenticated user can delete a team agent they don't belong to.

3. **Save errors silently discarded** — `let _ = store.save(...)` throughout `teams.rs`. On write failure, in-memory state diverges from disk and teams are lost on restart. At minimum, log the error.

### Minor issues

- `user_message` displays raw UUID as sender — should use email or display name
- `UserSearchInput` debounce timer not cleaned up on unmount
- Any team member can remove any other member including the creator
- Unused `put` import in `local_auth.rs`

### What looks good

- Auth token injection pattern (header for fetch, query param for EventSource) is clean
- AuthGate's synchronous token getter initialization avoids the race condition
- Team membership check on chat endpoint is correct
- SSE broadcast + buffer pattern for `user_message` follows existing conventions
- Tests for TeamStore roundtrip and filtering

---
_Automated review by Cthulu_

PR #123, opened 2026-03-21
feat: auth gate component + token injection

MPRR-Pandu:
## Summary
- `AuthGate.tsx`: login/signup UI, auth context, JWT expiry check, sync token init
- `client.ts`: token getter/setter, Bearer header on all `apiFetch` calls, 401 auto-logout, `withAuthToken()`/`getAuthTokenSync()` helpers for SSE

## Files (2)
- `cthulu-studio/src/components/AuthGate.tsx` (new, 176 lines)
- `cthulu-studio/src/api/client.ts` (+115 lines)

## Verification

### DIFF
--- a/cthulu-studio/src/api/client.ts
+++ b/cthulu-studio/src/api/client.ts
@@ -56,7 +56,9 @@ export function getAuthTokenSync(): string | null {
   return localStorage.getItem("cthulu_auth_token");
 }
 
-/** Append ?token= to a URL for EventSource connections (which can't set headers). */
+/** Append ?token= to a URL for EventSource connections (which can't set headers).
+ * Security note: tokens in query strings appear in server logs and browser history.
+ * Ensure server access logs redact query params in production. */
 export function withAuthToken(url: string): string {
   const token = getAuthTokenSync();
   if (!token) return url;
@@ -106,7 +108,7 @@ async function apiFetch<T>(
       if (res.status === 401 && getTokenFn) {
         localStorage.removeItem("cthulu_auth_token");
         window.location.reload();
-        return new Promise(() => {}) as T; // prevent downstream error handling during reload
+        throw new Error("Session expired");
       }
       const body = await res.text();
       log("error", `${method} ${path} -> ${res.status} (${elapsed}ms)`, body);
--- a/cthulu-studio/src/api/client.ts
+++ b/cthulu-studio/src/api/client.ts
@@ -106,6 +106,7 @@ async function apiFetch<T>(
       if (res.status === 401 && getTokenFn) {
         localStorage.removeItem("cthulu_auth_token");
         window.location.reload();
+        return new Promise(() => {}) as T; // prevent downstream error handling during reload
       }
       const body = await res.text();
       log("error", `${method} ${path} -> ${res.status} (${elapsed}ms)`, body);
--- a/cthulu-studio/src/api/client.ts
+++ b/cthulu-studio/src/api/client.ts
@@ -37,6 +37,39 @@ export function getServerUrl(): string {
   return getBaseUrl();
 }
 
+// ── Auth token injection ─────────────────────────────────────
+
+let getTokenFn: (() => Promise<string | null>) | null = null;
+
+/** Called by AuthGate to wire token getter into all API requests. */
+export function setAuthTokenGetter(fn: (() => Promise<string | null>) | null) {
+  getTokenFn = fn;
+}
+
+/** Get the current auth token (if available). Used by SSE EventSource callers. */
+export async function getAuthToken(): Promise<string | null> {
+  return getTokenFn ? getTokenFn() : null;
+}
+
+/** Get auth token synchronously from localStorage (for EventSource URLs). */
+export function getAuthTokenSync(): string | null {
+  return localStorage.getItem("cthulu_auth_token");
+}
+
+/** Append ?token= to a URL for EventSource connections (which can't set headers). */
+export function withAuthToken(url: string): string {
+  const token = getAuthTokenSync();
+  if (!token) return url;
+  const sep = url.includes("?") ? "&" : "?";
+  return `${url}${sep}token=${encodeURIComponent(token)}`;
+}
+
+/** Get auth headers for fetch-based streaming (non-apiFetch callers). */
+export async function getAuthHeaders(): Promise<Record<string, string>> {
+  const token = await getAuthToken();
+  return token ? { Authorization: `Bearer ${token}` } : {};
+}
+
 async function apiFetch<T>(
   path: string,
   options: RequestInit = {}
@@ -47,11 +80,21 @@ async function apiFetch<T>(
   log("http", `${method} ${path}`);
   const start = performance.now();
 
+  // Attach auth token if available
+  const authHeaders: Record<string, string> = {};
+  if (getTokenFn) {
+    const token = await getTokenFn();
+    if (token) {
+      authHeaders["Authorization"] = `Bearer ${token}`;
+    }
+  }
+
   try {
     const res = await fetch(url, {
       ...options,
       headers: {
         "Content-Type": "application/json",
+        ...authHeaders,
         ...options.headers,
       },
     });
@@ -59,6 +102,11 @@ async function apiFetch<T>(
     const elapsed = Math.round(performance.now() - start);
 
     if (!res.ok) {
+      // Auto-logout on 401 (expired/invalid token)
+      if (res.status === 401 && getTokenFn) {
+        localStorage.removeItem("cthulu_auth_token");
+        window.location.reload();
+      }
       const body = await res.text();
       log("error", `${method} ${path} -> ${res.status} (${elapsed}ms)`, body);
       throw new Error(`API error ${res.status}: ${body}`);
@@ -326,7 +374,7 @@ export function streamSessionLog(
   onLine: (line: string) => void,
   onDone: () => void
 ): () => void {
-  const url = `${getBaseUrl()}/api/agents/${agentId}/sessions/${sessionId}/stream`;
+  const url = withAuthToken(`${getBaseUrl()}/api/agents/${agentId}/sessions/${sessionId}/stream`);
   const source = new EventSource(url);
 
   source.addEventListener("line", (e: MessageEvent) => {
@@ -540,6 +588,7 @@ export async function createAgent(data: {
   permissions?: string[];
   append_system_prompt?: string | null;
   working_dir?: string | null;
+  team_id?: string | null;
 }): Promise<{ id: string }> {
   return apiFetch<{ id: string }>("/agents", {
     method: "POST",
@@ -586,7 +635,7 @@ export interface ResourceChangeEvent {
 export function subscribeToChanges(
   onEvent: (event: ResourceChangeEvent) => void
 ): () => void {
-  const url = `${getBaseUrl()}/api/changes`;
+  const url = withAuthToken(`${getBaseUrl()}/api/changes`);
   const source = new EventSource(url);
 
   const handler = (e: MessageEvent) => {
@@ -706,4 +755,68 @@ export async function getDashboardSummary(
     method: "POST",
     body: JSON.stringify({ channels }),
   });
+}
+
+// ── Profile ──────────────────────────────────────────────────
+
+export interface UserProfile {
+  id: string;
+  email: string;
+  name: string | null;
+  avatar_url: string | null;
+  created_at: string;
+}
+
+export async function getProfile(): Promise<UserProfile> {
+  return apiFetch<UserProfile>("/auth/me");
+}
+
+export async function updateProfile(data: { name?: string; avatar_url?: string }): Promise<UserProfile> {
+  return apiFetch<UserProfile>("/auth/me", { method: "PUT", body: JSON.stringify(data) });
+}
+
+export interface UserSearchResult {
+  id: string;
+  email: string;
+  name: string | null;
+}
+
+export async function searchUsers(query: string): Promise<{ users: UserSearchResult[] }> {
+  return apiFetch<{ users: UserSearchResult[] }>(`/auth/users/search?q=${encodeURIComponent(query)}`);
+}
+
+// ── Teams ────────────────────────────────────────────────────
+
+export interface TeamMember {
+  id: string;
+  email: string;
+  name: string | null;
+}
+
+export interface Team {
+  id: string;
+  name: string;
+  created_by: string;
+  members: TeamMember[] | string[];
+  created_at: string;
+}
+
+export async function listTeams(): Promise<{ teams: Team[] }> {
+  return apiFetch<{ teams: Team[] }>("/teams");
+}
+
+export async function createTeam(name: string): Promise<{ team: Team }> {
+  return apiFetch<{ team: Team }>("/teams", { method: "POST", body: JSON.stringify({ name }) });
+}
+
+export async function getTeam(id: string): Promise<{ team: Team }> {
+  return apiFetch<{ team: Team }>(`/teams/${id}`);
+}
+
+export async function addTeamMember(teamId: string, email: string): Promise<{ ok: boolean }> {
+  return apiFetch<{ ok: boolean }>(`/teams/${teamId}/members`, { method: "POST", body: JSON.stringify({ email }) });
+}
+
+export async function removeTeamMember(teamId: string, userId: string): Promise<{ ok: boolean }> {
+  return apiFetch<{ ok: boolean }>(`/teams/${teamId}/members/${userId}`, { method: "DELETE" });
 }
--- a/cthulu-studio/src/components/AuthGate.tsx
+++ b/cthulu-studio/src/components/AuthGate.tsx
@@ -59,7 +59,7 @@ export default function AuthGate({ children }: AuthGateProps) {
   const [token, setToken] = useState<string | null>(() => {
     const t = localStorage.getItem("cthulu_auth_token");
     if (t && isTokenValid(t)) {
-      setAuthTokenGetter(() => Promise.resolve(t));
+      setAuthTokenGetter(() => Promise.resolve(localStorage.getItem("cthulu_auth_token")));
       return t;
     }
     // Clear invalid/expired token
--- a/cthulu-studio/src/components/AuthGate.tsx
+++ b/cthulu-studio/src/components/AuthGate.tsx
@@ -1,0 +1,176 @@
+import { useState, useCallback, createContext, useContext } from "react";
+import { setAuthTokenGetter, getServerUrl } from "../api/client";
+
+// ── Auth context so any component can call logout ────────────
+
+interface AuthContextValue {
+  logout: () => void;
+  userEmail: string | null;
+}
+
+const AuthContext = createContext<AuthContextValue>({
+  logout: () => {},
+  userEmail: null,
+});
+
+/** Hook for any component to access logout + user info. */
+export function useAuth() {
+  return useContext(AuthContext);
+}
+
+interface AuthGateProps {
+  children: React.ReactNode;
+}
+
+/** Decode base64url to string (handles missing padding). */
+function b64urlDecode(str: string): string {
+  const padded = str.replace(/-/g, "+").replace(/_/g, "/") + "=".repeat((4 - (str.length % 4)) % 4);
+  return atob(padded);
+}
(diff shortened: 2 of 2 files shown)


saltyskip:

_Reviewing PR #123 — this may take a few minutes._

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  // Initialize token AND token getter synchronously to avoid race condition
+  // where children's effects fire API calls before the getter is wired up.
+  const [token, setToken] = useState<string | null>(() => {
+    const t = localStorage.getItem("cthulu_auth_token");
+    if (t && isTokenValid(t)) {
+      setAuthTokenGetter(() => Promise.resolve(t));
+      return t;
+    }
+    // Clear invalid/expired token
```
saltyskip, line 65: **Bug: stale token closure.** The token getter captures `t` by closure at login time. If the token is refreshed or rotated later, `setAuthTokenGetter(() => Promise.resolve(t))` will keep returning the original stale value forever.

Same issue on line 88 in `onSuccess`. Consider reading from localStorage each time instead:

```ts
setAuthTokenGetter(() => Promise.resolve(localStorage.getItem("cthulu_auth_token")));
```

This way `getTokenFn` always returns the current token.

cthulu-studio/src/api/client.ts
```diff
@@ -47,18 +80,33 @@ async function apiFetch<T>(
...
         ...options.headers,
       },
     });
 
     const elapsed = Math.round(performance.now() - start);
 
     if (!res.ok) {
+      // Auto-logout on 401 (expired/invalid token)
+      if (res.status === 401 && getTokenFn) {
+        localStorage.removeItem("cthulu_auth_token");
```
saltyskip, line 109: **Bug: 401 handler throws after reload.** `window.location.reload()` is async (the page doesn't stop executing immediately). The next line throws `new Error(...)`, which may trigger error handling in callers (toast notifications, error state) during the brief window before the page reloads.

Either `return` after the reload, or better yet, don't throw:

```ts
if (res.status === 401 && getTokenFn) {
  localStorage.removeItem("cthulu_auth_token");
  window.location.reload();
  // Return a never-resolving promise to prevent downstream error handling
  return new Promise(() => {}) as T;
}
```

cthulu-studio/src/api/client.ts
```diff
@@ -37,6 +37,39 @@ export function getServerUrl(): string {
...
+  return getTokenFn ? getTokenFn() : null;
+}
+
+/** Get auth token synchronously from localStorage (for EventSource URLs). */
+export function getAuthTokenSync(): string | null {
+  return localStorage.getItem("cthulu_auth_token");
+}
+
+/** Append ?token= to a URL for EventSource connections (which can't set headers). */
+export function withAuthToken(url: string): string {
```
saltyskip, line 62: **Security: token in query string.** `withAuthToken()` appends `?token=<jwt>` to EventSource URLs. Tokens in query strings get logged in server access logs, browser history, proxy logs, and any Referer headers. This is a known security anti-pattern for JWTs.

Since `EventSource` doesn't support custom headers natively, consider using `fetch` + `ReadableStream` (like `interactStream.ts` already does) with an `Authorization` header, or use a short-lived opaque ticket exchanged server-side.

cthulu-studio/src/api/client.ts
```diff
@@ -37,6 +37,39 @@ export function getServerUrl(): string {
...
+  getTokenFn = fn;
+}
+
+/** Get the current auth token (if available). Used by SSE EventSource callers. */
+export async function getAuthToken(): Promise<string | null> {
+  return getTokenFn ? getTokenFn() : null;
+}
+
+/** Get auth token synchronously from localStorage (for EventSource URLs). */
+export function getAuthTokenSync(): string | null {
```
saltyskip, line 55: Nit: `getAuthToken()` and `getAuthHeaders()` are exported but never used anywhere in this PR. The interact streams (`interactStream.ts`, `runStream.ts`) and the live SSE in `App.tsx` (line 168-169) don't use `withAuthToken` either — so those 3 EventSource connections will be unauthenticated when auth is enabled.

Either wire auth into those streams now, or add a TODO so it doesn't get forgotten.

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,176 @@
+import { useState, useCallback, createContext, useContext } from "react";
```
saltyskip, line 1: `AuthGate` is not imported or rendered anywhere (not in `App.tsx`, not in `main.tsx`). This component won't do anything until it's wired in. Is that intentional for a follow-up PR, or was the integration missed?

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,176 @@
...
+function parseToken(token: string): { email?: string; exp?: number } | null {
+  try {
+    const payload = token.split(".")[1];
+    return JSON.parse(b64urlDecode(payload));
+  } catch {
+    return null;
+  }
+}
+
+/** Check if token is valid: parseable, has email, not expired. */
```
saltyskip, line 41: **Missing signature verification.** `isTokenValid` only checks that the JWT is parseable and not expired — it doesn't verify the signature. A user could craft an arbitrary JWT with any email/exp and bypass the gate client-side.

This is fine *if* the server validates the signature on every API call (the real auth boundary). But worth noting that the client-side gate is purely cosmetic — make sure the backend enforces auth on all endpoints.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Blocking issues:**

1. **Stale token closure** — `setAuthTokenGetter(() => Promise.resolve(t))` captures the token value at login time. If the token is ever refreshed, all API calls will keep sending the old one. Read from localStorage instead.

2. **401 handler falls through to throw** — After `window.location.reload()`, execution continues and throws an error, which may flash error UI before the page reloads. Needs a `return` or never-resolving promise.

3. **Incomplete auth coverage** — Only 2 of 5 EventSource connections use `withAuthToken`. The SSE streams in `runStream.ts`, `interactStream.ts`, and `App.tsx` (flows/live) are unauthenticated. These will fail or leak data when auth is enabled.

4. **AuthGate not wired in** — The component isn't imported in `App.tsx` or `main.tsx`, so auth is effectively dead code.

**Non-blocking notes:**

- JWT tokens in query strings (`?token=...`) are a security concern (logged in server/proxy access logs, browser history, Referer headers). Consider using fetch+ReadableStream with Authorization headers for SSE, like `interactStream.ts` already does.
- Client-side `isTokenValid` doesn't verify the JWT signature — fine as long as the backend enforces auth on every endpoint.

---
_Automated review by Cthulu_

cthulu-studio/src/components/AuthGate.tsx
MPRR-Pandu: Fixed — token getter now reads from `localStorage.getItem()` each time instead of capturing by closure. Pushed.

cthulu-studio/src/api/client.ts
MPRR-Pandu: Fixed — returns `new Promise(() => {}) as T` after `window.location.reload()` to prevent downstream error handling. Pushed.

cthulu-studio/src/api/client.ts
MPRR-Pandu: Acknowledged — `EventSource` doesn't support headers, so `?token=` is the only option. The SSE wiring PR (next in chain) will use this. Server access logs should be configured to redact query params in production. A short-lived ticket system is a good future improvement.

cthulu-studio/src/api/client.ts
MPRR-Pandu: Intentional — `getAuthTokenSync()`, `withAuthToken()`, `getAuthHeaders()` are consumed in the next PR (#124 + SSE wiring PR). This PR adds the plumbing; follow-up PRs wire it into `interactStream.ts`, `runStream.ts`, etc.

cthulu-studio/src/components/AuthGate.tsx
MPRR-Pandu: Intentional — `AuthGate` is wired into `App.tsx` in PR #124 (the next PR). This PR is just the component + client plumbing.

cthulu-studio/src/components/AuthGate.tsx
MPRR-Pandu: Correct — client-side gate is cosmetic UX only. The real auth boundary is the `AuthUser` Axum extractor (PR #118, already merged) which verifies the JWT signature on every backend request. Client-side only checks expiry to avoid showing the app when the token is dead.

saltyskip:

_Re-reviewing PR #123 (previous HEAD: `760830a`, new HEAD: `96b4611`)_

cthulu-studio/src/api/client.ts
```diff
@@ -47,18 +80,34 @@ async function apiFetch<T>(
...
       },
     });
 
     const elapsed = Math.round(performance.now() - start);
 
     if (!res.ok) {
+      // Auto-logout on 401 (expired/invalid token)
+      if (res.status === 401 && getTokenFn) {
+        localStorage.removeItem("cthulu_auth_token");
+        window.location.reload();
```
saltyskip, line 110: **Bug: never-resolving promise leaks memory and skips `finally` blocks.**

`return new Promise(() => {}) as T` means any caller's `finally` block (e.g. `setLoading(false)`) will never run, and the promise stays in memory until the page reloads. Since you're already calling `window.location.reload()`, the practical impact is small, but if the reload is slow or fails (e.g. offline), the app is in a broken state.

Consider throwing instead and letting the reload race:
```ts
if (res.status === 401 && getTokenFn) {
  localStorage.removeItem("cthulu_auth_token");
  window.location.reload();
  throw new Error("Session expired");
}
```

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,176 @@
...
+      <AuthContext.Provider value={{ logout, userEmail: claims?.email ?? null }}>
+        {children}
+      </AuthContext.Provider>
+    );
+  }
+
+  return <AuthForm onSuccess={(t) => {
+    localStorage.setItem("cthulu_auth_token", t);
+    setAuthTokenGetter(() => Promise.resolve(t));
+    setToken(t);
```
saltyskip, line 91: **Bug: stale token captured in closure.**

`onSuccess` calls `setAuthTokenGetter(() => Promise.resolve(t))` which closes over the token string `t` from login time. Meanwhile `getAuthTokenSync()` reads live from localStorage. If the token is ever updated in localStorage (e.g. a refresh flow), the async getter returns the stale original while the sync getter returns the new one.

For consistency, use the same approach as the `useState` initializer:
```ts
setAuthTokenGetter(() => Promise.resolve(localStorage.getItem("cthulu_auth_token")));
```

cthulu-studio/src/api/client.ts
saltyskip, line 62: **Security: token passed in URL query string.**

Putting the JWT in `?token=` means it appears in server access logs, browser history, and can leak via `Referer` headers. This is a known EventSource limitation, but worth mitigating:
- Ensure the backend strips the token from access logs
- Consider short-lived tokens for SSE connections specifically
- At minimum, add a comment noting the security trade-off so future maintainers are aware

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,176 @@
...
+    setError("");
+    setLoading(true);
+
+    try {
+      const endpoint = mode === "signup" ? "/api/auth/signup" : "/api/auth/login";
+      const res = await fetch(`${getServerUrl()}${endpoint}`, {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ email, password }),
+      });
```
saltyskip, line 113: **Missing validation on received token.**

`onSuccess(data.token)` passes the token straight through without calling `isTokenValid()`. A malformed or already-expired token from the server would be stored and the user would appear logged in until the next API call 401s.

Add a check:
```ts
if (!data.token || !isTokenValid(data.token)) {
  setError("Received invalid token from server");
  return;
}
onSuccess(data.token);
```

saltyskip (changes requested):
## 🤖 Cthulu Review

### Blocking issues

1. **Incomplete SSE auth coverage** — The PR adds `withAuthToken()` to `streamSessionLog` and `subscribeToChanges`, but misses several other SSE/streaming endpoints that will fail when auth is enforced:
   - `runStream.ts:subscribeToRuns()` — uses bare `EventSource` without auth
   - `interactStream.ts:startInteract()` and `startNodeInteract()` — uses `fetch()` without auth headers
   - `App.tsx:168` — `new EventSource(url)` for `/api/flows/live` without auth

   These will all get 401s when `VITE_AUTH_ENABLED=true`. Either add `withAuthToken()` to all EventSource URLs and `getAuthHeaders()` to all fetch-based streams, or document which endpoints are intentionally unauthenticated.

2. **Stale token closure** — `onSuccess` captures the token string `t` in the auth getter closure. If the token is ever refreshed in localStorage, the async and sync getters will diverge. See inline comment.

3. **Never-resolving promise on 401** — `return new Promise(() => {}) as T` hangs forever if `window.location.reload()` is slow or fails. Throw instead. See inline comment.

### Non-blocking

- Token in URL query string for EventSource is a known limitation but should be documented with security implications.
- No `isTokenValid()` check on the token received from the server after login/signup.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #123 (previous HEAD: `96b4611`, new HEAD: `634571f`)_

cthulu-studio/src/api/client.ts
MPRR-Pandu: Fixed — throws `Error('Session expired')` instead of never-resolving promise.

cthulu-studio/src/components/AuthGate.tsx
MPRR-Pandu: Fixed — `onSuccess` now uses `localStorage.getItem()` consistently.

cthulu-studio/src/api/client.ts
```diff
@@ -47,18 +82,34 @@ async function apiFetch<T>(
...
         "Content-Type": "application/json",
+        ...authHeaders,
         ...options.headers,
       },
     });
 
     const elapsed = Math.round(performance.now() - start);
 
     if (!res.ok) {
+      // Auto-logout on 401 (expired/invalid token)
```
saltyskip, line 107: **Bug: Multiple concurrent 401s cause multiple reloads.**

If several requests are in-flight when the token expires, each one hits this branch independently — calling `removeItem` + `reload()` multiple times. This can cause race conditions and a visible flicker.

Add a guard:
```ts
let reloadingFor401 = false;
```
then:
```ts
if (res.status === 401 && getTokenFn && !reloadingFor401) {
  reloadingFor401 = true;
  localStorage.removeItem("cthulu_auth_token");
  window.location.reload();
  throw new Error("Session expired");
}
```

cthulu-studio/src/api/client.ts
```diff
@@ -326,7 +377,7 @@ export function streamSessionLog(
   onLine: (line: string) => void,
   onDone: () => void
 ): () => void {
-  const url = `${getBaseUrl()}/api/agents/${agentId}/sessions/${sessionId}/stream`;
+  const url = withAuthToken(`${getBaseUrl()}/api/agents/${agentId}/sessions/${sessionId}/stream`);
```
saltyskip, line 380: **Bug: Incomplete SSE auth coverage.** `streamSessionLog` and `subscribeToChanges` use `withAuthToken`, but three other SSE/streaming connections do NOT:

1. `subscribeToRuns` in `runStream.ts` (line 9: `EventSource` with no token)
2. `startInteract` / `startNodeInteract` in `interactStream.ts` (fetch calls with no `Authorization` header)
3. `App.tsx` line 168: the `/api/flows/live` SSE connection

When auth is enabled server-side, all of these will 401. Either apply `withAuthToken()` to every `EventSource` URL and add auth headers to the interact stream fetch calls, or export `getAuthHeaders()` and use it in `interactStream.ts`.

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,181 @@
+import { useState, useCallback, createContext, useContext } from "react";
```
saltyskip, line 1: **AuthGate is dead code** — it's not imported or rendered anywhere. `main.tsx` still renders `<App />` directly. For this PR to function, `main.tsx` or `App.tsx` needs to wrap the app:
```tsx
<AuthGate>
  <App />
</AuthGate>
```
Is the integration deferred intentionally? If so, note it in the PR description.

cthulu-studio/src/components/AuthGate.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const authEnabled = import.meta.env.VITE_AUTH_ENABLED === "true";
+
+  // Initialize token AND token getter synchronously to avoid race condition
+  // where children's effects fire API calls before the getter is wired up.
+  const [token, setToken] = useState<string | null>(() => {
+    const t = localStorage.getItem("cthulu_auth_token");
+    if (t && isTokenValid(t)) {
+      setAuthTokenGetter(() => Promise.resolve(localStorage.getItem("cthulu_auth_token")));
+      return t;
+    }
```
saltyskip, line 64: **Side effect in `useState` initializer.** Calling `setAuthTokenGetter()` inside the lazy initializer works today because it sets a module-level variable, but it's semantically a side effect in a place React expects to be pure. In Strict Mode (which this app uses), the initializer runs twice — the second call overwrites the first with the same value, so it's harmless here, but it's a fragile pattern.

Consider moving the `setAuthTokenGetter` call to a `useEffect` with an empty dependency array, or at minimum add a comment noting the Strict Mode double-invocation is safe.

cthulu-studio/src/api/client.ts
```diff
@@ -37,6 +37,41 @@ export function getServerUrl(): string {
...
+}
+
+/** Get auth token synchronously from localStorage (for EventSource URLs). */
+export function getAuthTokenSync(): string | null {
+  return localStorage.getItem("cthulu_auth_token");
+}
+
+/** Append ?token= to a URL for EventSource connections (which can't set headers).
+ * Security note: tokens in query strings appear in server logs and browser history.
+ * Ensure server access logs redact query params in production. */
```
saltyskip, line 61: **Nit: Token in query params leaks to server logs, Referer headers, and browser history.** The doc comment on `withAuthToken` already acknowledges this, which is good. Just noting that for the two `EventSource` call sites converted here, an alternative is to use `fetch` + `ReadableStream` (like `interactStream.ts` already does) so you can set the `Authorization` header instead. That would avoid the query-string exposure entirely.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Two blocking issues:**

1. **AuthGate is dead code** — not imported or rendered anywhere. `main.tsx` still renders `<App />` directly, so the auth gate never activates. Either integrate it in this PR or note in the description that integration is deferred.

2. **Incomplete SSE auth coverage** — only 2 of 5 SSE/streaming connections got auth tokens. `subscribeToRuns` (runStream.ts), `startInteract`/`startNodeInteract` (interactStream.ts), and the `/api/flows/live` EventSource in App.tsx will all 401 when auth is enabled server-side.

**Non-blocking:**
- Multiple concurrent 401 responses will each call `window.location.reload()` — add a guard flag
- `setAuthTokenGetter()` inside `useState` initializer is a side effect in a place React expects purity (runs twice in StrictMode)
- Token-in-query-string for EventSource could be avoided by switching to fetch+ReadableStream (like interactStream.ts already does)

---
_Automated review by Cthulu_

cthulu-studio/src/api/client.ts
MPRR-Pandu: Added security comment on `withAuthToken()` noting the query string trade-off.

cthulu-studio/src/components/AuthGate.tsx
MPRR-Pandu: Fixed — `AuthForm` now validates token with `isTokenValid()` before calling `onSuccess()`.

PR #126, opened 2026-03-23
feat: dashboard time range picker

MPRR-Pandu:
## Summary
Adds a dropdown to the dashboard to select time range: Today, Past 24h, Past 2 days, Past 7 days, Past 30 days.

The Python script already supported `--days N` — this just wires it up.

## Files (4)
- `handlers.rs` — accept `?range=` query param, map to `--days N` flag
- `client.ts` — pass range param to API
- `DashboardView.tsx` — range dropdown, auto-refresh on change
- `styles.css` — range select styling

## Verification
- `cargo check` — 0 errors

### DIFF
--- a/cthulu-studio/src/api/client.ts
+++ b/cthulu-studio/src/api/client.ts
@@ -695,8 +695,9 @@ export async function saveDashboardConfig(
   });
 }
 
-export async function getDashboardMessages(): Promise<DashboardMessages> {
-  return apiFetch<DashboardMessages>("/dashboard/messages");
+export async function getDashboardMessages(range?: string): Promise<DashboardMessages> {
+  const q = range && range !== "today" ? `?range=${range}` : "";
+  return apiFetch<DashboardMessages>(`/dashboard/messages${q}`);
 }
 
 export async function getDashboardSummary(
--- a/cthulu-studio/src/components/DashboardView.tsx
+++ b/cthulu-studio/src/components/DashboardView.tsx
@@ -79,6 +79,7 @@ export default function DashboardView() {
   const [showChannelDialog, setShowChannelDialog] = useState(false);
   const [channelInput, setChannelInput] = useState("");
   const [fetchedAt, setFetchedAt] = useState<string | null>(null);
+  const [range, setRange] = useState("today");
 
   // Summary state
   const [summaries, setSummaries] = useState<ChannelSummary[]>([]);
@@ -119,23 +120,22 @@ export default function DashboardView() {
     }
   }, []);
 
-  const loadMessages = useCallback(async () => {
+  const loadMessages = useCallback(async (r?: string) => {
     setLoading(true);
     setError(null);
     try {
-      const data = await getDashboardMessages();
+      const data = await getDashboardMessages(r ?? range);
       setChannels(data.channels || []);
       setFetchedAt(data.fetched_at || null);
     } catch (e) {
       const msg = (e as Error).message;
-      // Don't show error if just no channels configured
       if (!msg.includes("No channels configured")) {
         setError(msg);
       }
     } finally {
       setLoading(false);
     }
-  }, []);
+  }, [range]);
 
   const loadSummary = useCallback(async (channelData: SlackChannelMessages[]) => {
     if (channelData.length === 0) return;
@@ -212,7 +212,18 @@ export default function DashboardView() {
 
       <div className="dashboard-content">
         <div className="dashboard-section-header">
-          <h2>Today&apos;s Messages</h2>
+          <h2>Messages</h2>
+          <select
+            className="dashboard-range-select"
+            value={range}
+            onChange={(e) => { setRange(e.target.value); loadMessages(e.target.value); }}
+          >
+            <option value="today">Today</option>
+            <option value="1d">Past 24h</option>
+            <option value="2d">Past 2 days</option>
+            <option value="7d">Past 7 days</option>
+            <option value="30d">Past 30 days</option>
+          </select>
           <div className="dashboard-actions">
             {fetchedAt && (
               <span className="dashboard-fetched-at">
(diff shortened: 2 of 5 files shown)


saltyskip:

_Reviewing PR #126 — this may take a few minutes._

cthulu-studio/src/components/DashboardView.tsx
```diff
@@ -119,23 +120,22 @@ export default function DashboardView() {
...
       const msg = (e as Error).message;
-      // Don't show error if just no channels configured
       if (!msg.includes("No channels configured")) {
         setError(msg);
       }
     } finally {
       setLoading(false);
     }
-  }, []);
+  }, [range]);
```
saltyskip, line 138: **Bug: double-fetch on range change.** `loadMessages` now depends on `[range]`, so every time `setRange()` fires, `loadMessages` gets a new identity. This triggers the `useEffect` on line 154 (which depends on `loadMessages`), calling `loadMessages()` *again* — on top of the explicit `loadMessages(e.target.value)` in the `onChange` handler.

You have two options:

1. **Remove `range` from the dep array** — keep `[]` as it was. The `onChange` handler already passes the value explicitly via `loadMessages(e.target.value)`, so the callback doesn't need to close over `range`.

2. **Remove the explicit call in `onChange`** and instead use a `useEffect` that reacts to `range` changes. But option 1 is simpler and avoids the double-fetch.

cthulu-studio/src/api/client.ts
```diff
@@ -695,8 +695,9 @@ export async function saveDashboardConfig(
   });
 }
 
-export async function getDashboardMessages(): Promise<DashboardMessages> {
-  return apiFetch<DashboardMessages>("/dashboard/messages");
+export async function getDashboardMessages(range?: string): Promise<DashboardMessages> {
+  const q = range && range !== "today" ? `?range=${range}` : "";
```
saltyskip, line 699: Nit: `range` should be URL-encoded for correctness:
```ts
const q = range && range !== "today" ? `?range=${encodeURIComponent(range)}` : "";
```
The current values (`1d`, `2d`, etc.) are safe, but this is good hygiene if the value set ever expands.

saltyskip (changes requested):
## 🤖 Cthulu Review

**One bug to fix before merging:**

**Double-fetch on range change** — `loadMessages` depends on `[range]`, so changing the range recreates the callback, which triggers the `useEffect` that calls `loadMessages()` again — on top of the explicit `loadMessages(e.target.value)` in `onChange`. Every range change fires two API requests. Fix: revert the dep array back to `[]` since the `onChange` handler already passes the value explicitly.

The backend changes (match-based range mapping, no user input reaching the shell) and CSS are clean.

---
_Automated review by Cthulu_

saltyskip:

_Re-reviewing PR #126 (previous HEAD: `b834bdf`, new HEAD: `3b8b10c`)_

cthulu-studio/src/components/DashboardView.tsx
```diff
@@ -119,23 +120,22 @@ export default function DashboardView() {
...
       setFetchedAt(data.fetched_at || null);
     } catch (e) {
       const msg = (e as Error).message;
-      // Don't show error if just no channels configured
       if (!msg.includes("No channels configured")) {
         setError(msg);
       }
     } finally {
       setLoading(false);
     }
```
saltyskip, line 137: **Bug: double-fetch on range change.** `loadMessages` depends on `[range]`, and the mount `useEffect` depends on `[loadConfig, loadMessages]`. When the dropdown changes range:
1. `onChange` calls `setRange(newValue)` + `loadMessages(newValue)` — fires fetch #1
2. Re-render creates a new `loadMessages` ref (because `range` changed in its dep array)
3. `useEffect` sees `loadMessages` changed → re-runs → calls `loadConfig()` then `loadMessages()` — fires fetch #2

Every range change triggers two identical fetches plus an unnecessary config reload.

**Fix**: Remove `range` from the dependency array (the explicit argument `r` already handles it), or remove the explicit `loadMessages` call from `onChange` and let only the effect drive fetches.

scripts/slack_messages.py
```diff
@@ -141,7 +141,7 @@ def fetch(self, oldest: float, latest: float, conv_types: str,
             except (SlackApiError, RuntimeError) as e:
                 print(f"  Skipping {name}: {e}", file=sys.stderr)
                 continue
-            time.sleep(0.4)
+            time.sleep(0.2)
```
saltyskip, line 144: Halving the sleep from 0.4s → 0.2s is risky with the new 30-day range. `conversations.history` is Slack tier 2 (~20 req/min) and `conversations.replies` is tier 3 (~50 req/min). A 30d fetch across multiple active channels with many threaded messages will fire significantly more API calls — at 0.2s between calls you'll hit rate limits. The 0.4s value was already tight; consider keeping it, or at minimum adding retry-on-429 logic if you lower it.

cthulu-studio/src/api/client.ts
saltyskip, line 699: Nit: `range` isn't URL-encoded. Safe today since values come from a `<select>`, but if this function is ever called with user input it could break. Consider `encodeURIComponent(range)` or `new URLSearchParams({ range })`.

saltyskip (changes requested):
## 🤖 Cthulu Review

**Blocking: double-fetch bug on range change.** The `loadMessages` callback includes `range` in its dependency array, which causes the mount `useEffect` to re-fire on every dropdown change — resulting in two identical Slack API fetches and a redundant config reload per interaction. This needs to be fixed before merge.

**Concern: rate-limit sleep halved with larger time ranges.** The 0.4s → 0.2s sleep reduction in `slack_messages.py` combined with the new 30-day range (which fetches far more messages/threads) is likely to trigger Slack rate limits. Either keep the original sleep, add retry-on-429 handling, or at minimum test a 30d fetch on a busy workspace before merging.

The Rust backend changes (query param extraction, match dispatch, timeout bump) look clean. The CSS and overall UX approach are solid.

---

_Automated review by Cthulu_

